PR #2, opened 2020-05-06
Add a margin to ensure the timeline is visible

jacobwillemsma:
Current timeline is clipped on at least standard Macbook Pro display (although - seems it would be on any display). This pushes it over to not conflict with other text.

Note: Might have mobile considerations, but not sure how many people will be doing DApp development on mobile quite yet 😛 

**Current**:
image

**New**:
image


austintgriffith:
Maybe I could just have the providers down in the bottom right stack up better so they aren't in the way.

jacobwillemsma:
Also could work!

PR #37, opened 2020-08-09
Contract refactor

TomAFrench:
Rather than using `useContractLoader` twice, we now connect the signer onto the contract when passing it into the `FunctionForm` component.

Also linting fixes.


TomAFrench:
This might warrant a comment in the code explaining why we do this but I'll leave that up to you.

PR #45, opened 2020-09-02
Chainlink example

PatrickAlphaC:
Hi team!

In this example pretty much all the instructions for the user are in `ChainlinkHints.jsx`

The only other bit that I did was I changed the default provider for rinkeby to be an environment variable... it doesn't quite go with your standard, I'm wondering if I should fix it?

Chainlink doesn't work great only locally, so you don't have to run your own chain here. I've been meaning to write a "fund-contracts" script, but my ethers.js is still a bit new. Any thoughts? Let me know if stuff is confusing!


codenamejason:
@PatrickAlphaC let me know if I can close this. There are already a few chainlink branches, have a look and see if they are up to par with what you were trying to do.

PR #46, opened 2020-09-04
React canvas object as a global var

dhalsim:
This change makes it easy to play with react-canvas-draw from JS console.

Here is a snippet how that works:

```
drawingCanvas.current.points = [{x:0, y:0}, {x:drawingCanvas.current.canvas.drawing.width, y:drawingCanvas.current.canvas.drawing.height}]

drawingCanvas.current.saveLine({
    brushColor: drawingCanvas.current.props.brushColor,
    brushRadius: drawingCanvas.current.props.brushRadius
})

drawingCanvas.current.simulateDrawingLines({
    lines: drawingCanvas.current.lines,
    immediate: true
})
```

And the result:

https://nifty.ink/QmZBxNgPEWS3oQhuqHqxm6dM8iH8YbUreqSzcNGiEz1hmV

I tested it with my local front-end and xdai network

### DIFF
--- a/packages/react-app/src/InkCanvas.js
+++ b/packages/react-app/src/InkCanvas.js
@@ -71,6 +71,7 @@ export default function InkCanvas(props) {
       }catch(e){console.log("Drawing Error:",e)}
     }
   }
+  window.drawingCanvas = drawingCanvas
   showDrawing()
 }, [props.ipfsHash])
 
--- a/packages/react-app/src/InkCanvas.js
+++ b/packages/react-app/src/InkCanvas.js
@@ -45,6 +45,7 @@ export default function InkCanvas(props) {
           }
         }
     }
+    window.drawingCanvas = drawingCanvas
     loadPage()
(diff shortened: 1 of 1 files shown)


packages/react-app/src/InkCanvas.js
```diff
@@ -71,6 +71,7 @@ export default function InkCanvas(props) {
       }catch(e){console.log("Drawing Error:",e)}
     }
   }
+  window.drawingCanvas = drawingCanvas
```
azf20, line 74: Nice! I think this could go here, or alternatively above loadPage().

azf20 (approved):
@austintgriffith I think this is good to merge?

azf20:
note that the example code breaks the undo button, I think due to double saving (when the simulateDrawingLines is invoked), but that is not related to the code change itself. Merging

azf20 (approved):
LGTM

PR #49, opened 2020-09-07
Add 'Fill background' and 'Draw frame' functionality

ososco:
One-click button to fill background according to picked color, and another button to draw instant frames according to picked color and radius.

image

### DIFF
--- a/packages/react-app/src/InkCanvas.js
+++ b/packages/react-app/src/InkCanvas.js
@@ -256,21 +256,66 @@ const onFinishFailed = errorInfo => {
   console.log('Failed:', errorInfo);
 };
 
+const triggerOnChange = (lines) => {
+  let saved = JSON.stringify({
+    lines: lines,
+    width: drawingCanvas.current.props.canvasWidth,
+    height: drawingCanvas.current.props.canvasHeight
+  });
+
+  drawingCanvas.current.loadSaveData(saved, true);
+  drawingCanvas.current.lines = lines;
+};
+
+const undo = () => {
+  if (!drawingCanvas.current.lines.length) return;
+
+  if (drawingCanvas.current.lines[drawingCanvas.current.lines.length - 1].ref) {
+    drawingCanvas.current.lines[0].brushColor = drawingCanvas.current.lines[drawingCanvas.current.lines.length - 1].brushColor;
+    let lines = drawingCanvas.current.lines.slice(0, -1);
+    triggerOnChange(lines);
+  } else {
+    let lines = drawingCanvas.current.lines.slice(0, -1);
+    triggerOnChange(lines);
+  }
+};
+
 const fillBackground = (color) => {
   let width = drawingCanvas.current.props.canvasWidth;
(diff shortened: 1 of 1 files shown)


packages/react-app/src/InkCanvas.js
```diff
@@ -256,7 +256,45 @@ const onFinishFailed = errorInfo => {
...
+      { x: 0, y: 0 },
+      { x: width, y: height }
+    ]
+  });
+
+  let saved = drawingCanvas.current.getSaveData();
+  drawingCanvas.current.loadSaveData(saved, true);
+};
 
+const drawFrame = (color, radius) => {
```
azf20, line 276: this is great

packages/react-app/src/InkCanvas.js
```diff
@@ -256,7 +256,45 @@ const onFinishFailed = errorInfo => {
   console.log('Failed:', errorInfo);
 };
 
+const fillBackground = (color) => {
```
azf20, line 259: Nice. I wonder if there is a way to inject this so it's a true background (currently it's definitely a fill, but it kind of fills in everything)

packages/react-app/src/InkCanvas.js
```diff
@@ -353,6 +391,18 @@ if (props.mode === "edit") {
           />
         </Col>
     </Row>
+    <Row style={{ width: "90vmin", margin: "0 auto", marginTop:"4vh", justifyContent:'center'}}>
+        <Col span={4}>
```
azf20, line 395: I got a bit of overlap between these buttons, I think you can remove the Col wrapper (it's not doing too much)
image

packages/react-app/src/InkCanvas.js
```diff
@@ -353,6 +391,18 @@ if (props.mode === "edit") {
...
+          <Button
+          onClick={() => fillBackground(color)}
+          ><BgColorsOutlined />Background</Button>
```
azf20, line 398: Should we maybe call this "Fill" rather than background?

azf20 (commented):
Really nice! Simple elegant implementation. Couple of quick comments

packages/react-app/src/InkCanvas.js
dhalsim: maybe that could be possible if we forked react-canvas-draw and changed every code from

```
ctx.clearRect(0, 0, ctx.canvas.width, ctx.canvas.height);
```

to fill the color we passed (there is no option to pass either)

```
ctx.fillStyle = FILL_COLOR;
ctx.fillRect(0, 0, ctx.canvas.width, ctx.canvas.height);
```

I think Osama's approach is better, very nice workaround.

azf20: I guess I meant more to show up behind any drawing that has already been done. But there is a lot of complexity there I think

dhalsim: Actually there can be an easy way to make this always stay on the background. Maybe I can check it this evening if that works. basically if we manage to always draw background before everything else that would solve our problem

PR #50, opened 2020-09-07
show static hint image to setup metamask xdai network

dhalsim:
I copied hint image from burner-wallet and wanted to show to the user in case metamask network is not xdai.
we already show a modal message so I put it there, and added some style to position it on the corner. 

it is a bit weird: appears on center and than goes to the corner. It was quick and dirty but I'm open to suggestions.


codenamejason:
@azf20 this is old, should I merge it?

PR #51, opened 2020-09-08
CLR

veronicaz41:
In this version, there's the concept of Round.
- Owner can only add recipients before the round starts. Once the round starts, the recipients set is locked. 
- The owner can start the Round and config the Round duration.
- The Round lasts some time which gives contributors time to consider the recipients set and contribute to the matching pool. During the Round, anyone can contribute to the matching pool. Allowed donors can donate to specific projects.
- Once the Round duration reached, the Round is closed. Matching can be calculated then.

Once the Round starts, we cannot add new recipients to the recipients set. New recipients might need to wait for other Rounds, rather than start accepting donations immediately. This is not ideal. But we are not limited to have only one Round at any given time. We can deploy this CLR contract again, add new recipients and start another Round. A CLR factory would be useful in this case.


packages/buidler/contracts/CLR.sol
```diff
@@ -0,0 +1,162 @@
...
+            );
+        }
+
+        calculatedToIndex = toIndex;
+        if (calculatedToIndex >= recipients.length) {
+            emit MatchingCalculated();
+        }
+    }
+
+    function withdraw(uint16 index) public isRoundClosed returns (uint256) {
```
veronicaz41, line 139: rename this withdraw to "recipientWithdraw". Have a new withdraw for admin to withdraw what's left.

packages/buidler/contracts/CLR.sol
```diff
@@ -0,0 +1,162 @@
...
+        // matching = matchingPool * matchingWeight / totalMatchingWeight
+        uint256 matchingWeight;
+        bool withdrawn;
+    }
+
+    Recipient[] public recipients;
+
+    uint256 public matchingPool;
+
+    mapping(address => bool) public donorAllowList;
```
veronicaz41, line 28: maybe worth to put this functionality in another contract

packages/buidler/contracts/CLR.sol
```diff
@@ -0,0 +1,162 @@
...
+        });
+        recipients.push(r);
+
+        uint256 index = recipients.length - 1;
+        emit RecipientAdded(addr, data, index);
+        return index;
+    }
+
+    function allowDonor(address donor) public onlyOwner {
+        donorAllowList[donor] = true;
```
veronicaz41, line 96: Add a event

packages/buidler/contracts/CLR.sol
```diff
@@ -0,0 +1,162 @@
...
+    modifier isRoundClosed() {
+        require(
+            roundStart != 0 &&
+                getBlockTimestamp() >= roundStart.add(roundDuration),
+            "Round is not closed"
+        );
+        _;
+    }
+
+    function startRound(uint256 _roundDuration) public onlyOwner {
```
veronicaz41, line 64: We should not be able to start round multiple times

PR #55, opened 2020-09-15
CLR v0.001

veronicaz41:
Admin page
1. transfer owner
2. add recipient
3. allow donor, block donor
4. start round (set round duration)
5. trigger calculateMatching

Recipient page
1. submit project info
2. show the decision - whether gets added to the recipient set
3. trigger calculateMatching (optional)
4. see how much donation the recipient gets, withdraw

Donor page
1. view recipients projects
2. register to be a donor, gets added to allow list.
3. see how much donation to a project will affect the money the project eventually gets (optional)
4. donate (vote) to projects directly
5. send money to matching pool
6. see the matching result, how much money each recipient project gets

### DIFF
--- a/packages/buidler/scripts/deploy.js
+++ b/packages/buidler/scripts/deploy.js
@@ -58,7 +58,7 @@ async function main() {
   // custom deploy (to use deployed addresses dynamically for example:)
   const donorManager = await deploy("DonorManager");
   const clr = await deploy("CLR", [donorManager.address]);
-  await clr.transferOwnership("0x025645A569b3e60F803bFFC88f0E2e38b7526B3d");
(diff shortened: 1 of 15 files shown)


packages/buidler/scripts/deploy.js
```diff
@@ -38,14 +38,14 @@ function readArgumentsFile(contractName) {
 async function autoDeploy() {
   const contractList = fs.readdirSync(config.paths.sources);
   return contractList
-    .filter((fileName) => isSolidity(fileName))
```
veronicaz41, line 41: my linter did all this

packages/react-app/src/views/Debug.jsx
```diff
@@ -0,0 +1,221 @@
+import React, { useCallback, useEffect, useState } from "react";
```
veronicaz41, line 1: The original App.jsx

PR #57, opened 2020-09-30
Instantly run yarn start, chain & deploy in one command

easonchai:
Using concurrently,
Should solve some points @austintgriffith mentioned in #7


pabloruiz55:
I gave this a try but I'm not convinced it helps. You end up with the react server and the hardhat network both running and printing output on the same terminal window which is messy. And you still need a second terminal to keep running yarn deploy as you make changes to the contracts.

Screen Shot 2021-06-25 at 5 20 59 PM

PR #64, opened 2020-10-14
Fix dependency issue

### LINKED ISSUE #63: [Bug] caniuse-lite is outdated
On clean install getting this error for `yarn start` for the `nifty-ink-dev` branch

```
Starting the development server...

Browserslist: caniuse-lite is outdated. Please run:
npx browserslist@latest --update-db
```

Had this issue as well in https://github.com/austintgriffith/eth.build/pull/41. I was able to solve it there by switching from yarn to npm. Problem is `create-eth-app` [cannot use npm](https://github.com/PaulRBerg/create-eth-app#creating-an-app) since it relies on yarn workspaces.

Until then the only fix is to run the command above

pi0neerpat:
Updates yarn.lock to fix outdated dependency
closes #63

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -43,7 +43,29 @@
     ts-invariant "^0.4.4"
     tslib "^1.10.0"
 
-"@apollo/react-hooks@^3.1.3":
+"@apollo/react-components@^3.1.5":
+  version "3.1.5"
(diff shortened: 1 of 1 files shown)


pabloruiz55:
There's been a lot of changes to the master branch and the nifty-ink-dev is 400+ commits behind now. Probably makes sense to close this PR. @azf20

codenamejason:
I would have to agree @pabloruiz55

Issue #66, opened 2020-10-31
Buidler -> Hardhat?


jamespfarrell:
[Builder has become hardhat?](https://medium.com/nomic-labs-blog/buidler-has-evolved-introducing-hardhat-4bccd13bc931)

Wondering if I should wait for scaffold-eth to upgrade, if that's going to happen?

jamespfarrell:
Austin told me on twitter, that it will upgraded at some point.

aspiers:
I guess this can be upgraded now that #68 is merged?

azf20:
Yep :)

PR #68, opened 2020-11-07
Buidler -> Hardhat

azf20:
- Replaces Buidler with Hardhat per the [migration instructions](https://hardhat.org/guides/migrate-from-buidler.html)
- Removes the /contracts package (not currently used)

Notes:
- This is a straight-swap change, there should be no change in functionality
- Hardhat deploy saves the YourContract.json in a different place (updated Publish script accordingly)
- The local 8545 chainId seems to be different (I had to update my metamask config)

(cc: #66)


jamespfarrell:
Thank you both so much!

PR #70, opened 2020-11-15
Fix eslint warnings

aspiers:
This fixes a bunch of eslint warnings which were cluttering up the console and making it very hard to see any new issues being introduced.

### DIFF
--- a/packages/react-app/src/views/Subgraph.jsx
+++ b/packages/react-app/src/views/Subgraph.jsx
@@ -1,3 +1,5 @@
+/* eslint-disable jsx-a11y/accessible-emoji */
+
 import React, { useState, useEffect } from "react";
 import "antd/dist/antd.css";
 import { MailOutlined } from "@ant-design/icons";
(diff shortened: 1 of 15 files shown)


aspiers:
There's still one unresolved warning in the `usePoller` hook due to a missing dependency on `fn`.  I haven't been able to fix it without causing that hook to enter an infinite spiral of doom which kills any browser tab it comes into contact with.  My understanding of React hooks is buried under 6 months of dust and cobwebs but I was wondering if it would maybe make sense to use `useCallback` instead of `useRef` - however that's a wild guess so take with a large pinch of salt.

packages/react-app/src/views/Subgraph.jsx
```diff
@@ -76,8 +76,8 @@ function Subgraph(props) {
           <div style={{ margin: 32, marginTop: 32, margin: "auto" }}>
             You will find that parsing/tracking events with the <span style={highlight}>useEventListener</span> hook becomes a chore for every new project.
           </div>
-          <div style={{ margin: 32, marginTop: 32, margin: "auto" }}>
-            Instead, you can use <a href="https://thegraph.com/docs/introduction" target="_blank">The Graph</a> with 🏗 scaffold-eth by following these steps:
+          <div style={{ marginTop: 32, margin: "auto" }}>
```
aspiers, line 79: Whoops, this change was supposed to go in the commit fixing duplicate margins.

azf20 (approved):
This looks good, thanks for cleaning up!

PR #76, opened 2020-12-05
dark mode enabled

pavlovcik:
if your iOS/MacOS device has dark mode enabled, it'll make the UI dark


azf20:
Hey thanks! Though conscious the colours on this one can get a bit spacey for the default view (my mac is dark mode), @austintgriffith maybe we can add a dark mode toggle to scaffold-eth to save Pavlov's eyes?

PR #78, opened 2020-12-05
Fix yarn test and make it easier to find

aspiers:
Fix `yarn test` to run on the `localhost` hardhat network, and mention it in the quickstart in the README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,14 @@ yarn deploy
 
 ```
 
+> in a fourth terminal window:
+
+```bash
+
+yarn test
+
+```
+
 🔏 Edit your smart contract `YourContract.sol` in `packages/hardhat/contracts`
 
 📝 Edit your frontend `App.jsx` in `packages/react-app/src`
--- a/packages/hardhat/package.json
+++ b/packages/hardhat/package.json
@@ -25,7 +25,7 @@
   },
   "scripts": {
     "chain": "hardhat node",
-    "test": "hardhat test --network hardhatevm",
+    "test": "hardhat test",
     "compile": "hardhat compile",
     "deploy": "hardhat run scripts/deploy.js && hardhat run scripts/publish.js",
     "watch": "node scripts/watch.js",


README.md
```diff
@@ -40,6 +40,14 @@ yarn deploy
 
 ```
 
+> in a fourth terminal window:
+
+```bash
+
+yarn test
```
azf20, line 47: Test doesn't do much here as part of the core setup, but it is definitely something cool to mention! Maybe this could go further down (as having it here implies that it is part of the basic lift-off)

packages/hardhat/package.json
```diff
@@ -25,7 +25,7 @@
   },
   "scripts": {
     "chain": "hardhat node",
-    "test": "hardhat test --network hardhatevm",
```
azf20, line 28: Nice!

azf20 (changes requested):
Nice! I think the README update may make sense a bit further down (rather than as part of the core setup) cc @austintgriffith

README.md
aspiers: Well, for those of us who subscribe to the TDD philosophy, it *is* part of the basic lift-off ;-)  But I appreciate that there are others who don't.  It is already mentioned further down but it's buried in a long list under [the "Original Quickstart with TODO List" section](Original Quickstart with TODO List) which is doubly hidden because it's an `<h2>` nested inside `<h1>built with 🏗 scaffold-eth:</h1>`.  I assume this was unintentional and either this h2 should be an h1, or this h1 should be an h2, so that either way they are peer sections not parent/child.  But this points to a wider issue - IMHO the existing README is overwhelmingly long making it hard to follow the structure, and as such should be split into smaller pieces.  Please let me know how you want to proceed and I'll adjust this PR accordingly.

BTW I would have just dropped this contentious commit from the PR for the sake of getting the `--network hardhatevm` brokenness fixed ASAP, but then due to [bad design of GitHub's review system](https://github.com/isaacs/github/issues/998) this discussion thread would vanish.

PR #79, opened 2020-12-05
Add a mechanism for cleanly supporting interface files

aspiers:
Currently `deploy.js` and `publish.js` automatically try to deploy and publish any `.sol` files they can find.  However interface files do not need to be deployed, and attempting to do so results in this error:

    NomicLabsHardhatPluginError: You are trying to create a contract
    factory for the contract YourInterface, which is abstract and
    can't be deployed.

So add a mechanism for skipping deployment and publishing, and make it easy to do this by adding interface files to `hardhat.config.js`.  Later we may want to consider automatically detecting interface files, but this is a simple solution for now.

### DIFF
--- a/packages/hardhat/contracts/YourContract.sol
+++ b/packages/hardhat/contracts/YourContract.sol
@@ -2,8 +2,9 @@ pragma solidity >=0.6.0 <0.7.0;
 
 import "hardhat/console.sol";
 //import "@openzeppelin/contracts/access/Ownable.sol"; //https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable.sol
+import "./YourInterface.sol";
 
-contract YourContract {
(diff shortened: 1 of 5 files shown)


jamespfarrell:
Good man Adam!

azf20:
Hey thanks! This introduces some complexity into the core branch. I think you can achieve the same outcome (no warnings during publish + deploy) by putting interfaces in a folder within contracts? This is also bit more legible for the user. What do you think?
image

aspiers:
Interesting idea, I'll try it.

pabloruiz55:
@aspiers did @azf20 suggestion work? If yes, I'll close this PR.
Also, we recently updated Scaffold-eth to work with Hardhat Deploy, so probably things are a bit different now.

aspiers:
Yeah, probably no longer valid due to the migration to Hardhat.

PR #80, opened 2020-12-05
Add SPDX-License-Identifier to fix remix warning

aspiers:
When editing in remix via remixd, this gives a clean compile.


azf20:
Thanks! @austintgriffith do you feel strongly about whether some other SPDX license would make more sense https://docs.soliditylang.org/en/v0.6.8/layout-of-source-files.html

pabloruiz55:
In recent versions we have
SPDX-License-Identifier: MIT

PR #90, opened 2021-01-12
Added re-enterancy and attacker contract examples

Divide-By-0:
Edit: See updated and cleaned version here: https://github.com/austintgriffith/scaffold-eth/tree/reentrancy-example

The re-enterancy contract is vulnerable through its poorly written withdraw() function. Withdraw_safe() is not. To test the attack do this:

Player 1 (usually an incognito tab):
In the re-enterancy contract, Deposit 0.001 ETH (converted into GWEI by clicking *)
Verify the amount in the contract is $0.001 (or whatever the price of ETH is)

Player 2 (attacker):
In the attacker contract, deposit 0.001 ETH as well.
Verify the contract now has $0.002
From the attacker contract, withdraw
Notice the balance of the attacker contract now has 2x the funds it should, and re-enterancy has 0. Also, the Player 1's balance in the contract seems to still be 0.001.

The takeaway? Follow the checks-effects-interaction pattern in all your functions :)

At least for now, this PR isn't meant to be merged! Just for future reference.

### DIFF
--- a/packages/hardhat/contracts/Attacker.sol
+++ b/packages/hardhat/contracts/Attacker.sol
@@ -1,0 +1,37 @@
+pragma solidity >=0.6.0 <0.7.0;
+
+import "./Reenterancy.sol";
+
+contract Attacker {
+    // event SetPurpose(address sender, string purpose);
+    // event Withdraw(address sender, uint256 amount);
+    Reenterancy reenterancy;
+    bool hasAttacked;
+
+    constructor(address payable contractAddress) public { 
+        // hasAttacked = false;
+        reenterancy = Reenterancy(contractAddress);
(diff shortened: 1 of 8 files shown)


codenamejason:
@Divide-By-0 should create a new branch for this.

pabloruiz55:
Hey @Divide-By-0 nice example! 
We should not merge this to master, but we could add the sample to the tutorials section. Would you like to make any additional changes to this code or is it ok to be published as-is?
We would close the PR, but leave the branch open for people to checkout.

Divide-By-0:
Yes, a tutorial section sounds great -- OK with everything you suggested @pabloruiz55.

PR #97, opened 2021-01-20
Add autotest capability

NotRealAI:
This change makes it so that using the command `yarn run autotest` a watcher will be started which will automatically run tests in a particular file upon changes. This allows for faster iteration on tests, especially in conjunction with using `it.only`. Inspired by testing using Guard in Ruby.

This pull request uses a slightly modified version of the plugin `hardhat-watcher` which is waiting on approval for merge in that repository.

[https://github.com/NotRealAI/hardhat-watcher](https://github.com/NotRealAI/hardhat-watcher)

### DIFF
--- a/README.md
+++ b/README.md
@@ -390,6 +390,11 @@ yarn run watch
 yarn run test
 ```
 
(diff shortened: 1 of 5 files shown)


NotRealAI:
Autotest capability has been merged into hardhat-watcher, so this PR should be remade with the main repo

PR #103, opened 2021-01-31
Dotenv hardhat

wschwab:


codenamejason:
@wschwab I would just comment out the networks using mnemonic and not delete. Gives user the option to use a key or mnemonic.

pabloruiz55:
I like this change as the most common mistake people will probably make is committing changes to the hardhat config file, exposing their Infura api key. Same thing could be done to constants.js file in the React folder which exposes Infura, Etherscan and Blocknative api keys.

wschwab:
> @wschwab I would just comment out the networks using mnemonic and not delete. Gives user the option to use a key or mnemonic.

I could do that - a compromise is very reasonable. I'll need to re-merge master in (Scaffold moves fast, I'm sure this is out of date already), but I'll try to put it in the queue for the weekend

pabloruiz55:
Cool! You might want for #192 to be merged which brings a few changes I believe.

pabloruiz55:
After taking a second look I'd say that moving the api keys to a .env file is great, but I'd keep the networks using the mnemonic file instead of changing it to use a pk in the .env file. To keep it in line with the `hardhat generate` deployment workflow.

pabloruiz55:
I'd also use this opportunity to move the other API keys to this file like Blocknative's and Etherscan's.

kenonchain:
This is super useful. Any update?

codenamejason (commented):
I think it is super useful as well. I still think we should allow for the mnemonic with the option for using the pk from env. This would be great for the mainnet-master branch.

codenamejason (approved):
this is a good start @StErMi and we can work from this on the env vars.

StErMi:
> this is a good start @StErMi and we can work from this on the env vars.

yeah I was thinking about something like that. for `hardhart` we need the dotenv, for `react` we can just create a `.env` with `REACT_APP_` variables. 

we `could` but it's a lot of effort to also create a task that add/replace a `.env` var called `BURN_WALLET_PRIVATE_KEY` if we really want to keep using burning wallet for deployment

PR #119, opened 2021-02-26
Improved UI + Updated polling times + reduced Contract Reader calls

amogh-jrules:


azf20:
@amogh-jrules is this the latest to review?

amogh-jrules:
@azf20 Not anymore 😅I'll push a few more commits and then tag you :)

PR #121, opened 2021-02-28
feat: add gas estimation skipping ability and fork note for tenderly

ironsoul0:
Here is the minimal contract function that would show that the transaction hash is not shown. As a consequence, we will not be able to export it to Tenderly.

```
function getDivision(uint256 x, uint256 y) public returns (uint256) {
  uint256 result = x / y;
  map[msg.sender] = result;
   return result;
}
```

Pass `y` as 0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -790,7 +790,23 @@ tenderly export <transactionHash>
 ```
 Which will upload them to tenderly.co/dashboard!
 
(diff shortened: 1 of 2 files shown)


azf20 (approved):
Great stuff @ironsoul0 !

PR #131, opened 2021-03-12
Adding styles to the navigation, adding homepage

ChangoMan:


packages/react-app/src/NftyWallet.js
```diff
@@ -95,427 +93,193 @@ export default function NftyWallet(props) {
...
+                ipfsConfigInfura={ipfsConfigInfura}
+                gasPrice={props.gasPrice}
+                calculatedVmin={calculatedVmin}
+                transactionConfig={transactionConfig}
+              />
+            </Route>
+
+            {/* <Route path="/create">
+            <div>
+              <CreateInk
```
ChangoMan, line 189: Is this route and component still needed? If not, we should delete it to make sure it doesn't compile.

packages/react-app/src/components/Navigation.js
```diff
@@ -0,0 +1,37 @@
...
+import { NavLink } from 'react-router-dom';
+import '../styles/navigation.css';
+
+function Navigation({ address }) {
+  return (
+    <header className="site-nav">
+      <Row align="middle" justify="space-between">
+        <Col xs={24} lg={12}>
+          <div className="logo">
+            <NavLink to="/">Logo</NavLink>
```
ChangoMan, line 12: Do you guys have a logo?

ChangoMan:
I was getting a bunch of warnings about unused imports, so I went through some of the files and removed unused imports. I'm pretty sure this will help speed up compile and reduce final package size.

viraj124:
we probably would need to remove the fake data before merge cc @austintgriffith @azf20 @ironsoul0 @sadda11asm

packages/react-app/src/components/Navigation.js
ironsoul0: I think it can be just `xNFT` for now.

packages/react-app/src/NftyWallet.js
ironsoul0: Yeap. We do not need it. We left it for debugging purposes only.

packages/react-app/src/NftyWallet.js
```diff
@@ -95,427 +93,193 @@ export default function NftyWallet(props) {
...
+                injectedProvider={props.injectedProvider}
+              />
+            </Route>
+            <Route path="/debug">
+              <DebugContracts {...props} />
+            </Route>
+            <Route path="/recent">
+              <AllInks />
+            </Route>
+            {/* <Route path="/curated">
```
ironsoul0, line 141: We actually want to have this page in production.

ironsoul0: UPD: I see that CuratedInks is used on the HomePage. My apologies.

ironsoul0:
LGTM. We can merge.

packages/react-app/src/NftyWallet.js
viraj124: @ironsoul0 should we remove the fake data after merge then ?

PR #145, opened 2021-03-23
Solidity to v0.8.2 and graph-node fix

rhsr:
These are the changes I found myself making to get the project into a state that I wanted to work off. 

1) Pinning `graphprotocol/graph-node` to `v0.22.0`. This eliminates [the CLI issue](https://github.com/graphprotocol/graph-node/issues/2271). It seems that the graph team is aware of it and has pushed a fix but I am still getting the error when trying to start a node using a clean clone of scaffold. I decided that this would make it easier for new users as it seems that there isn't much gained from pointing at the latest version for the target audience of this repo.

2) Add solidity `0.8.2` compiler to hardhat. Based on the breaking changes from `0.7.x` -> `0.8.x`, I feel that a new user would be best set up for success by building habits in `0.8.x`. Things like explicitly casting global address vars as payable could be non-obvious to new developers when they eventually find themselves in `0.8.x` and beyond. I decided to keep `0.7.6` in the config as to not break any solidity built that is specifically scoped to `<=0.7.6`, and as an opportunity to demonstrate the power of hardhat and multiple compiler versions. I've updated the README to reflect this.

I'm a novice with solidity so please let me know if I've missed anything obvious.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ Look for the [HardHat](https://hardhat.org) console.log() output in the `yarn ch
 ![image](https://user-images.githubusercontent.com/2653167/93150263-dae25180-f6b5-11ea-94e1-b24ab2a63fa5.png)
 
 
-🚽 UNDERFLOW!?! (🚑 [Solidity >0.8.0](https://docs.soliditylang.org/en/v0.8.0/) catchs this [with Checked Arithmetic](https://blog.soliditylang.org/2020/12/16/solidity-v0.8.0-release-announcement/)!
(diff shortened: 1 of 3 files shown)


codenamejason:
@rhsr should we change 8.2 to 8.4 now?

pabloruiz55:
In recent commits we've already upgraded to Solc v.0.8.4, so we can drop that part of the PR.
@azf20 what are your thoughts on the Pinning graphprotocol/graph-node to v0.22.0? Is that still relevant?

codenamejason (approved):

codenamejason:
@azf20 please see Pablo's comment above.

PR #153, opened 2021-04-03
Merkel Root Logic Integration

viraj124:
- Modified the current version with Merkel root logic where the Merkel root consists of the artwork hashes, so when a buyer tries to claim the proof is verified before minting


codenamejason:
I like the added value this PR adds.

azf20:
@viraj124 is this still live?

viraj124:
> @viraj124 is this still live?

@azf20 this pr can be deleted since this https://github.com/austintgriffith/scaffold-eth/tree/merkle-root-buyer-mints branch was already shared by the community

PR #157, opened 2021-04-10
Good tokens

BarakChamo:


codenamejason:
Why don't we create a new branch called good-token for this?

codenamejason:
I think we should create a good-tokens branch for this and not merge into master. What do you think @azf20 ?

azf20:
Agreed @codenamejason - @BarakChamo did you mean to PR this into the master branch?

Issue #129, opened 2021-03-06
Unhandled Rejection (Error): missing response


daniel234x:
Please how can I solve this error, I keep getting after I run `Yarn start` 

`Unhandled Rejection (Error): missing response (requestBody="{\"method\":\"eth_accounts\",\"params\":[],\"id\":45,\"jsonrpc\":\"2.0\"}", requestMethod="POST", serverError={}, url="http://localhost:8545", code=SERVER_ERROR, version=web/5.0.9)`

What should I do ?

ccarnino:
I was experiencing this error too. The issue was that the correct order is to `yarn chain` before `yarn start`. So when the frontend starts if finds the node already running.

hustqmk:
I find out that open 127.0.0.1:3000 in my vitual box is ok, but will meet the same error on my host machine

PR #160, opened 2021-04-19
Fixed the quickstart terminal commands order

ccarnino:
Inverted the order on which the node and the react app is start.

Currently the react app was started first and it was throwing an error, because it was not finding the node running yet. As per described in [Issue 129](https://github.com/austintgriffith/scaffold-eth/issues/129).

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,19 +59,8 @@ required: [Node](https://nodejs.org/dist/latest-v12.x/) plus [Yarn](https://clas
 
 ```bash
 git clone https://github.com/austintgriffith/scaffold-eth.git
-
 cd scaffold-eth
-```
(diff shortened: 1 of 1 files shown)


codenamejason:
We can't merge this yet, yarn install needs to be ran first. Otherwise the chain won't run.

codenamejason:
@ccarnino did you want to fix this?

ccarnino:
Hey @codenamejason I understood your feedback and it made total sense. I made the suggested change and now you have yarn install as fist step.

codenamejason (approved):
I think this is good to go

ccarnino:
Great!

PR #167, opened 2021-04-27
polling interval (.env) for infura rpc calls

harryranakl:
polling interval (.env) for infura rpc calls. it will help reduce infura calls and draining infura request limit. updated balance and gasprice hooks to have polling interval.


codenamejason:
This has been a huge issue in the community with the polling and would like to merge this.

azf20:
I pushed https://github.com/austintgriffith/scaffold-eth/pull/170 which significantly reduces polling root cause. Please try the latest master

PR #175, opened 2021-05-12
PowVT-etherscan-specify-contract

PowVT:


azf20:
Nice! Is it worth adding `npx hardhat clean` to any of the package.json commands? 🤔

PowVT:
> Nice! Is it worth adding `npx hardhat clean` to any of the package.json commands? 🤔

For sure! I took a stab at it let me know if that looks right.

codenamejason (approved):

PR #176, opened 2021-05-12
Hunter signatorio

ChangoMan:
Adding some styles and making it responsive


packages/react-app/src/App.jsx
```diff
@@ -1,40 +1,61 @@
...
+      package: WalletConnectProvider, // required
+      options: {
+        infuraId: INFURA_ID,
+      },
+    },
+  },
+});
 
-function App(props) {
+const logoutOfWeb3Modal = async () => {
```
ChangoMan, line 49: Moved these up here cause my linter was complaining about them being down below

azf20 (approved):
Looks good to me, thanks so much! I think key UX question is the vertical space on desktop

PR #178, opened 2021-05-14
add new background and logo

ChangoMan:
I cut out the logo from the old background and put it in separately, so we can make it smaller on desktop.


azf20 (approved):
Lovely stuff (and thanks for the linting!)

PR #197, opened 2021-06-13
chore: create new useLatestRef helper hook

YPAzevedo:
This PR adds a new helper hook to help keep track of non-primitive values and keep a stable reference, while still remembering the latest value


codenamejason:
what do you think @austintgriffith

PR #209, opened 2021-06-21
Hardhat 2.4.0 and Custom Errors sample

pabloruiz55:
- Upgrades to Hardhat 2.4.0
- Adds example in YourContract.sol on how to define a custom error w/params (new in Solidity 0.8.4) and how to revert with it.
- Updates transactor to show error data if it exists.

*Note: Waffle/Chai does not yet support custom errors so I didn't add a test for it yet.

Screen Shot 2021-06-19 at 4 57 16 PM

Screen Shot 2021-06-19 at 4 49 27 PM

### DIFF
--- a/packages/hardhat/contracts/YourContract.sol
+++ b/packages/hardhat/contracts/YourContract.sol
@@ -1,4 +1,4 @@
-pragma solidity >=0.6.0 <0.9.0;
+pragma solidity >=0.8.0 <0.9.0;
 //SPDX-License-Identifier: MIT
 
 import "hardhat/console.sol";
@@ -10,14 +10,22 @@ contract YourContract {
 
   string public purpose = "Building Unstoppable Apps";
 
+  error EmptyPurposeError(uint code, string message);
+
   constructor() {
     // what should we do on deploy?
   }
 
   function setPurpose(string memory newPurpose) public {
-    purpose = newPurpose;
-    console.log(msg.sender,"set purpose to",purpose);
-    emit SetPurpose(msg.sender, purpose);
-  }
+      if(bytes(newPurpose).length == 0){
+          revert EmptyPurposeError({
+              code: 1,
+              message: "Purpose can not be empty"
+          });
+      }
 
+      purpose = newPurpose;
+      console.log(msg.sender,"set purpose to",purpose);
+      emit SetPurpose(msg.sender, purpose);
+  }
 }
--- a/packages/react-app/src/helpers/Transactor.js
+++ b/packages/react-app/src/helpers/Transactor.js
@@ -115,7 +115,7 @@ export default function Transactor(providerOrSigner, gasPrice, etherscan) {
         console.log("Transaction Error:", e.message);
         notification.error({
           message: "Transaction Error",
-          description: e.message,
+          description: e.data ? e.data.message : e.message,
         });
         if (callback && typeof callback === "function") {
           callback(e);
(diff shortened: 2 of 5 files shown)


pabloruiz55:
@azf20 I'd love to get this merged. Please let me know if you have some time to give it a try. Wanna make sure nothing in the workflow is broken with the update to hardhat 2.4.0
TY!

packages/react-app/src/helpers/Transactor.js
```diff
@@ -115,7 +115,7 @@ export default function Transactor(providerOrSigner, gasPrice, etherscan) {
         console.log("Transaction Error:", e.message);
         notification.error({
           message: "Transaction Error",
-          description: e.message,
+          description: e.data ? e.data.message : e.message,
```
azf20, line 118: This logic works for Metamask, but I think this may need additional handling for the burnerSigner - I see this: 
image

Essentially the error thrown seems to be different. I got it showing up nicely with the burner by looking for `e.reason`, i.e.
```
notification.error({
          message: "Transaction Error",
          description: e.data ? e.data.message : e.reason ? e.reason : e.message,
        });```
though that's a bit gnarly!
I tried with WalletConnect via Metamask mobile, which didn't actually throw at all, so maybe there is something required on that side.

**Review by azf20 — CHANGES_REQUESTED** [review]
Nice! I think just needs slightly different handling to show the error for a burnerSigner as well as Metamask!

File: packages/hardhat/contracts/YourContract.sol [hunk 2]
```diff
@@ -10,14 +10,22 @@ contract YourContract {
...
     // what should we do on deploy?
   }
 
   function setPurpose(string memory newPurpose) public {
-    purpose = newPurpose;
-    console.log(msg.sender,"set purpose to",purpose);
-    emit SetPurpose(msg.sender, purpose);
-  }
+      if(bytes(newPurpose).length == 0){
+          revert EmptyPurposeError({
```
azf20, line 21: This will also break the subgraph per this issue, which will hopefully be resolved soon https://github.com/graphprotocol/graph-node/issues/2577

packages/react-app/src/helpers/Transactor.js
pabloruiz55: Thanks @azf20 while I agree it looks bad, this issue is not something that has been introduced by this PR. If on the master branch you do the same you will see the same result if the function reverts. Check this (on master branch):
```
  function setPurpose(string memory newPurpose) public {
    require (bytes(newPurpose).length > 0, "Purpose is empty!");
    purpose = newPurpose;
    console.log(msg.sender,"set purpose to",purpose);
    emit SetPurpose(msg.sender, purpose);
  }
```

**With BurnerWallet:**
Screen Shot 2021-06-25 at 9 08 44 PM

**With Metamask:**
Screen Shot 2021-06-25 at 9 08 32 PM

So, I wouldn't stop this PR because of this, since we are still getting the same behavior, but possible open another issue to get it fixed. Although, I've been looking at this for a while and can't find a way to get that error correctly parsed, which seems to be the issue (when using the burnerWallet the error is not parsed correctly by ethers.js?)

azf20: Sure I think the snippet I shared in the comment parses the error correctly for the BurnerSigner:
```
notification.error({
          message: "Transaction Error",
          description: e.data ? e.data.message : e.reason ? e.reason : e.message,
        });
```
I'll push to this branch and you can verify?

azf20 (approved):
Added a fix for parsing burner errors, LGTM if you are happy!

pabloruiz55:
Merged. There's issue https://github.com/austintgriffith/scaffold-eth/issues/224
which causes us not to fully benefit from this when using a BurnerWallet instead of Metamask.

Issue #198, opened 2021-06-14
Scaffold-eth React-app should gracefully handle not having an internet connection


pabloruiz55:
We should be able to develop in airplane mode / Chrome offline simulation while connected to a local chain.
Instead, we get the following error upon page load. 

<img width="1414" alt="Screen Shot 2021-06-14 at 1 11 12 PM" src="https://user-images.githubusercontent.com/526558/121924438-470be780-cd12-11eb-9088-7f76aa920d3e.png">

First step should be to handle the error so the app still manages to load.

Second step should be reviewing all the components the basic template has so we remove the internet connectivity dependency. I.E: When connecting to local chain the default template still tries to connect to mainnet rpc node.

TaldenV:
I'm on a road trip and would love to see this feature implemented. Because I'm going to be offline anyway, I'll be able to spend some time looking into handling the errors so the app can still load. I created and pulled a branch, but decided I should probably be commenting on the issue before trying to do a pull request.

pabloruiz55:
> I'm on a road trip and would love to see this feature implemented. Because I'm going to be offline anyway, I'll be able to spend some time looking into handling the errors so the app can still load. I created and pulled a branch, but decided I should probably be commenting on the issue before trying to do a pull request.

Awesome! Let me know when you have something and I'll test it out.

TaldenV:
Ok, I was able to find some fixes. It looks like the main things that are causing the connection issues to pop up are the gas hook, the faucet, and the mainnet connection in app.jsx. I was able to comment out all of the connection issues. It might be due to getting rid of the debug warnings, but once you get rid of the warnings you can access and edit the dapp just fine. Once in airplane mode, you just need to add the hardhat RPC network to the wallet, then add a hardhat account pre-loaded with 100 eth. The RPC network and accounts can be found in the terminal immediately after running the yarn chain.

One issue I found is that the hardhat accounts don’t reset in the metamask wallet once the chain is closed and reopened. It seems to be an issue that the hardhat community is aware of/working on (Reference: https://github.com/MetaMask/metamask-extension/issues/10290).

[...]

PR #210, opened 2021-06-22
Sanity checks for when connection is offline

pabloruiz55:
A stab at solving #198.

With these fixes, if you are offline and you still want to code while on localhost + local chain, you should not get any errors from the components that are looking for stuff on mainnet.

It doesn't solve the issue when you go to Chrome's console -> network -> simulate offline, 
BUT it does work fine when you are truly offline or in airplane mode.

### DIFF
--- a/packages/react-app/src/App.jsx
+++ b/packages/react-app/src/App.jsx
@@ -55,8 +55,8 @@ if (DEBUG) console.log("📡 Connecting to Mainnet Ethereum");
 //
 // attempt to connect to our own scaffold eth rpc and if that fails fall back to infura...
 // Using StaticJsonRpcProvider as the chainId won't change see https://github.com/ethers-io/ethers.js/issues/901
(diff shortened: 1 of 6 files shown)


azf20 (changes requested):
Nice! Works great offline but when I went online it wasn't connecting - not sure if it's a browser specific thing, but I think `navigator.onLine` is a boolean rather than a string when I print it out?
`navigator.onLine: false`
I think `navigator.onLine === 'online'` can be replaced by just checking for `navigator.onLine`?

pabloruiz55:
> Nice! Works great offline but when I went online it wasn't connecting - not sure if it's a browser specific thing, but I think `navigator.onLine` is a boolean rather than a string when I print it out?
> `navigator.onLine: false`
> I think `navigator.onLine === 'online'` can be replaced by just checking for `navigator.onLine`?

You are absolutely right. Not sure why I did that there and why I thought it was working fine, but it's fixed now. Thanks!

NhlanhlaHasane (approved):

pabloruiz55:
@azf20 whenever you get a chance please review again and approve the changes so Github allows me to merge. Thanks!

azf20 (approved):
LGTM!

PR #212, opened 2021-06-23
Added support for hardhat/evm network calls

pabloruiz55:
Added a new menu to the front-end so we can easily interact with Hardhat network stuff like advancing time on the local chain.

This PR adds:
- evm_increaseTime
- evm_setNextBlockTimestamp
- hardhat_setBalance

https://user-images.githubusercontent.com/526558/123113809-38f64f00-d415-11eb-95d0-b660d2a93085.mov

Screen Shot 2021-07-01 at 5 03 59 PM


packages/react-app/src/App.jsx
```diff
@@ -512,6 +512,14 @@ function App(props) {
               Support
             </Button>
           </Col>
+          {localProvider._network && localProvider._network.chainId == 31337 && (
```
ShravanSunder, line 515: There's a proliferation of the checks for the hardhat chain in local network all over App.jsx.

I would suggest this pattern should be modified and there should a variable on top of app.jsx that replaces all these checks to make it easier for new devs to understand what is happening.  
```
const isHardHatNetwork = localProvider._network && localProvider._network.chainId == 31337
```
and replacing the 4-6 instances of the check in this file with that constant

packages/react-app/src/components/HardhatEVM.jsx
```diff
@@ -0,0 +1,175 @@
...
+import {
+  useOnBlock,
+} from "../hooks";
+import { formatEther, parseEther } from "@ethersproject/units";
+import { parse } from "graphql";
+
+const { Option } = Select;
+const { utils, BigNumber } = require("ethers");
+
+/*
```
ShravanSunder, line 17: for new components, i would suggest we use JS doc format.  will help future documentation, and hints for new developers.

below is an example from a hook
```
/**
 * Checks whether a contract exists on the blockchain, returns true if it exists, otherwise false
 * 
  ~ Features ~
  - Provide contractAddress to check if the contract is deployed
  - Change provider to check contract address on different chains (ex. mainnetProvider)
 * @param provider 
 * @param contractAddress 
 * @returns 
 */
```

packages/react-app/src/components/HardhatEVM.jsx
```diff
@@ -0,0 +1,175 @@
...
+  ~ Features ~
+*/
+
+export default function HardhatEVM(props) {
+  const [address, setAddress] = useState();
+  const [visible, setVisible] = useState(false);
+  const [chainTimestamp, setChainTimestamp] = useState();
+  const [accountBalance, setAccountBalance] = useState();
+
+  const localProvider = props.localProvider;
```
ShravanSunder, line 37: there doesn't seem to be a reason to only destructure one of the prop vars

packages/react-app/src/components/HardhatEVM.jsx
```diff
@@ -0,0 +1,175 @@
...
+    const aNum = parseInt(parseEther(form["hardhat_setBalance_amount"]));
+    console.log(aNum.toString(16));
+    await localProvider.send("hardhat_setBalance", [form["hardhat_setBalance_address"], "0x"+aNum.toString(16)]);
+    await localProvider.send("evm_mine");
+    const newBalance = await localProvider.getBalance(form["hardhat_setBalance_address"]);
+    setAccountBalance(formatEther(newBalance));
+  }
+
+  const [ form, setForm ] = useState({})
+  const setField = (field, value) => {
```
ShravanSunder, line 81: the recommended react method is the use callback of setState function, rather than using the useState value

```
setForm(oldValue => { ...oldValue, [field]:value
});
```

ShravanSunder:
@pabloruiz55  left some comments

packages/react-app/src/components/HardhatEVM.jsx
pabloruiz55: What would be the right way to do it? :)

ShravanSunder: Oh what I meant was we should destructure al the variables or none of them and use props directly. It’s inconsistent to do half way

pabloruiz55: ah, gotcha, thanks!

PR #127, opened 2021-03-05
Add typescript + smart contract types with hardhat-typechain

0xGabi:
Note the typescript flag `strict` on `tsconfig.json` was set to false to reduce the number of changes.

PR #223, opened 2021-06-26
Created typescript versions of the app and hardhat folder

ShravanSunder:
I've created a branch of scaffold eth based on master that runs on Typescript and vite and react. 
- The hardhat deployment/publish/ etc files are also in TS

Notes:
- i refactored most of the components to typescript
- the views are also in typescript
- it has the components from buidlGuidlv2 feature branch
- currently, all the views open
- since its based of vite instead of webpack, its much faster to load (vite runs on esbuild), it takes about 5 seconds from a cold start in my laptop

Format:
- There's 2 new package folders called hartdhat-ts and vite-app-ts that have the ts version
- I didn't modify the react-app and hardhat folders

Has the latest changes from the hardhat-deploy additionally.


pabloruiz55:
Wow thanks @ShravanSunder! This is super appreciated. Please hold on until someone has the chance to give this a proper read as it's a huge change.
AFAICT, one could merge this all but the ts version is separate and people could keep using everything asi-is if they don't know/care about typescript, right?

I'm a bit worried about code duplication as with this we would need to maintain both and any change has to also be applied to the ts versions.

ShravanSunder:
> Wow thanks @ShravanSunder! This is super appreciated. Please hold on until someone has the chance to give this a proper read as it's a huge change.
> AFAICT, one could merge this all but the ts version is separate and people could keep using everything asi-is if they don't know/care about typescript, right?
> 
> I'm a bit worried about code duplication as with this we would need to maintain both and any change has to also be applied to the ts versions.


Yes it’s different folders and I didn’t my best not to impact the existing code.  it’s a bit of duplication. The next steps after this would be to reduce that by:
- to work on pulling out and making al the hooks compatible with both versions. Via eth-hooks package (I’m looking into this with Calvin)
- Move the components into eth-components package to reduce the code duplication and use them in both
- Create any more common packages if need to further reduce duplication. 

The vite app uses vite instead of webpack and is much faster to build and deploy than webpack. And I’ve updated hardhat ts to use hardhat deploy and a easier to understand folder structure. 
-

pabloruiz55:
@ShravanSunder have you seen PR #223 ? Anything you can bring in from there?

ShravanSunder:
> @ShravanSunder have you seen PR #223 ? Anything you can bring in from there?

223 is this issue so I’m a bit confused 🤔

pabloruiz55:
Sorry, wrong one. I meant #127

ShravanSunder:
> Sorry, wrong one. I meant #127 

This is already done in hardhat-ts package!  The generated types are loaded in the vite-app-ts as a local file module. 😃

pabloruiz55:
Awesome, I'll close that PR then in favor of this more comprehensive one.

ShravanSunder:
I've merged the latest master into this

NhlanhlaHasane (approved):

ShravanSunder:
> Awesome, I'll close that PR then in favor of this more comprehensive one.

@NhlanhlaHasane  seems to have approved my changes, i think i've merged the latest master from sunday night.  @pabloruiz55 what would the next steps be? =)

pabloruiz55:
I'd like to get @austintgriffith , @codenamejason @azf20 's input first maybe since it's such a big PR.

austintgriffith:
Let's not merge this into master. Instead, let's create a new branch for this. When we make big change to master let's create a staging branch and show it off and have some folks use it before going to master.

ShravanSunder:
> Let's not merge this into master. Instead, let's create a new branch for this. When we make big change to master let's create a staging branch and show it off and have some folks use it before going to master.

I've updated the target branch of merge to `typescript` @pabloruiz55  @austintgriffith

ShravanSunder:
As per conversation on wednesday, merging this into the typescript branch in scaffold-eth so devs can experiment with this branch.

PR #231, opened 2021-06-29
Fix: react warnings

harryranakl:


pabloruiz55:
Maybe dumb question as I'm not an expert in React. Is there a way to fix those warnings instead of having them ignored by linter?

harryranakl:
nt sure, nt the best way to fix the warnings

packages/react-app/src/components/ThemeSwitch.jsx
```diff
@@ -5,6 +5,7 @@ import { useThemeSwitcher } from "react-css-theme-switcher";
 export default function ThemeSwitcher() {
   const theme = window.localStorage.getItem("theme");
   const [isDarkMode, setIsDarkMode] = useState(!(!theme || theme === "light"));
+  // eslint-disable-next-line
```
pabloruiz55, line 8: Instead of `eslint-disable-next-line` you can remove the status variable since the warning is `Line 8:35:  'status' is assigned a value but never used  no-unused-vars`.

pabloruiz55:
Since the majority of this code has been written @austintgriffith @azf20 I believe, please chime in if you are ok with this solution. 

Warnings make CI like Netlify's fail deployment for example, so it would be good to get rid of all of these one way or another.

azf20:
Thanks! Probably worth addressing them, I tried fixing here https://github.com/austintgriffith/scaffold-eth/pull/250

pabloruiz55:
Dropping in favor of #250

Issue #224, opened 2021-06-27
When using ethers.js wallet, Transactor is unable to properly parse errors


pabloruiz55:
How to reproduce:

1. Deploy the default contracts and UI
2. Use the app with a BurnerWallet (which is the case by default, make sure you are not using Metamask).
3. Submit an empty new purpose, which will revert with "purpose can not be empty".
4. Instead of seeing a nice error, the error is caught but it's unable to properly parse it and can't access all the data.

You might, instead, get something like this:
```
[...]

PR #237, opened 2021-06-30
Find the error message when using ethers js signer

pabloruiz55:
See #224

### DIFF
--- a/packages/react-app/src/helpers/Transactor.js
+++ b/packages/react-app/src/helpers/Transactor.js
@@ -118,16 +118,10 @@ export default function Transactor(providerOrSigner, gasPrice, etherscan) {
 
         return result;
       } catch (e) {
-        let message;
-        if(network.chainId == 31337)
-        {
-          if(e.data) // Metamask signer returns data 
-            message = e.data.message;
-          else // Ethers js needs parsing the error
-            message = JSON.parse(JSON.parse(JSON.stringify(e.error)).body).error.message;
-        }else{
-          message = e.data ? e.data : JSON.stringify(e);
-        }
+        if (DEBUG) console.log(e);
+        // Accounts for Metamask and default signer on all networks
+        let message = e.data && e.data.message ? e.data.message : e.error && JSON.parse(JSON.stringify(e.error)).body ? JSON.parse(JSON.parse(JSON.stringify(e.error)).body).error.message : e.data ? e.data : JSON.stringify(e);
+
         console.log("Transaction Error:", message);
         notification.error({
           message: "Transaction Error",
--- a/packages/react-app/src/helpers/Transactor.js
+++ b/packages/react-app/src/helpers/Transactor.js
@@ -119,12 +119,16 @@ export default function Transactor(providerOrSigner, gasPrice, etherscan) {
         return result;
       } catch (e) {
         let message;
-        if(e.data) // Metamask signer returns data 
-          message = e.data.message;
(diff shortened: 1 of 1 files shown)


packages/react-app/src/helpers/Transactor.js
```diff
@@ -118,11 +118,16 @@ export default function Transactor(providerOrSigner, gasPrice, etherscan) {
 
         return result;
       } catch (e) {
-        console.log(e);
-        console.log("Transaction Error:", e.message);
+        let message;
+        if(e.data) // Metamask signer returns data 
+          message = e.data.message;
+        else // Ethers js needs parsing the error
+          message = JSON.parse(JSON.parse(JSON.stringify(e.error)).body).error.message;
```
azf20, line 125: This throws on Rinkeby, seems like the error object has a different shape 
image
Maybe if it doesn't have this expected shape we can just print the whole slug? (as I feel like there might be a few variations, based on type of error message...)

azf20 (changes requested):
Works nicely on hardhat! This threw on Rinkeby as the returned error shape is different, maybe try it and set message=slug in the catch?

azf20 (approved):

packages/react-app/src/helpers/Transactor.js
azf20: My bad this is still valid, I had switched back to Master

azf20 (changes requested):
Verified that this is:
Good on hardhat
Ok (without the good descriptor) on Kovan
Throws on Rinkeby

pabloruiz55:
New attempt, tested on localhost, ropsten, kovan, rinkeby and mainnet.
All show pretty much a raw error (as I can't find a proper way to perfectly parse each method on each network, and even then they don't expose the revert reason), except localhost that can show the revert reason.

localhost-mm
ropsten-wallet
localhost-wallet
ropsten-mm
mainnet-mm
mainnet-wallet
rinkeby-mm
rinkeby-wallet
kovan-mm
kovan-wallet

packages/react-app/src/helpers/Transactor.js
```diff
@@ -118,16 +118,10 @@ export default function Transactor(providerOrSigner, gasPrice, etherscan) {
...
-          if(e.data) // Metamask signer returns data 
-            message = e.data.message;
-          else // Ethers js needs parsing the error
-            message = JSON.parse(JSON.parse(JSON.stringify(e.error)).body).error.message;
-        }else{
-          message = e.data ? e.data : JSON.stringify(e);
-        }
+        if (DEBUG) console.log(e);
+        // Accounts for Metamask and default signer on all networks
+        let message = e.data && e.data.message ? e.data.message : e.error && JSON.parse(JSON.stringify(e.error)).body ? JSON.parse(JSON.parse(JSON.stringify(e.error)).body).error.message : e.data ? e.data : JSON.stringify(e);
```
pabloruiz55, line 123: I like this better, still not great, but at least it eliminates the chainId-based conditional.

azf20: Nice one!

azf20 (approved):

azf20:
As you say it's not perfect but great stuff testing all those networks, and as you say this error seems to be down to some combination of the network clients / ethers.js / metamask and it wasn't introduced by this original PR - created https://github.com/austintgriffith/scaffold-eth/issues/241 to track

PR #247, opened 2021-07-04
Update balance when provider / address changes

azf20:
Balances are only updated when there is a new block. This is especially strange during local development where you don't get new blocks on an ongoing basis. This change updates the balance when the provider or address changes


feynman1668 (commented):
It is working fine after this commit

pabloruiz55 (approved):

PR #263, opened 2021-07-13
Adding GSN support to Scaffold-ETH

drortirosh:
This PR adds GSN support for scaffold-eth: the "injected" provider is wrapped by GSN, and the sample "YourContract" adds the required modifications to support meta-transactions.


azf20:
This is great! Interested to see that the OZ library is the recommended way in Solidity, nice. And some good devex too (starting up with the chain)

codenamejason:
@drortirosh is this still being worked on?

codenamejason:
@drortirosh I am going to close this pr

Issue #258, opened 2021-07-09
Missing Wallet Icon?


austintgriffith:
I was doing a live demo and noticed the Wallet icon is missing now on master

It should look like: 
![image](https://user-images.githubusercontent.com/2653167/125102125-32fca100-e098-11eb-9b19-9e2e859002be.png)

Issue #265, opened 2021-07-13
When you click the wallet icon for the faucet it just spins


austintgriffith:
![image](https://user-images.githubusercontent.com/2653167/125474237-ca212fa4-60b2-452e-a8ff-4aae378a8852.png)

Something happened to the wallets or the provider is wrong here?

austintgriffith:
https://github.com/austintgriffith/scaffold-eth/pull/271

PR #271, opened 2021-07-14
fix faucet and account wallet

azf20:
To fix https://github.com/austintgriffith/scaffold-eth/issues/258 and https://github.com/austintgriffith/scaffold-eth/issues/265


austintgriffith (approved):
yesssssss

PR #272, opened 2021-07-14
MvpNFT

MercuricChloride:
A simple and easily human readable contract for a 1/1 NFT that stores a string at the top of the contract.


codenamejason:
don't we want this to be a branch?

PR #273, opened 2021-07-16
Add xDAI to MetaMask button

ososco:
Added a button in help section, where users can add xDAI with one-click on MetaMask instead of following manual instructions.

Manual instructions removed from help section, external like with instruction on xDAI chain added instead.

&nbsp;

image


azf20:
This one escalated... LGTM!

PR #281, opened 2021-07-20
Fixed Infinite loop bug in useContractReader hook

pratyaksh123:
If the contract function that we are calling has no arguments and we provide a pollTime, then tried state variable is never set to true, this results in an infinite loop of contract reads.


azf20:
aha nice catch! I think I also saw this bug last night https://github.com/austintgriffith/scaffold-eth/pull/282

pratyaksh123:
Oh cool, i will close this then.

PR #282, opened 2021-07-20
Small bugs

azf20:
- setTried in contractReader
- React to changes in the customAddresses in the contract loader
- Use a URL that works for WalletConnect


austintgriffith (approved):
🙏

PR #286, opened 2021-07-22
Update ContractReader.js

korotkih:
Added setTried(true) in case tried === false and updateValue() function was called to prevent calling updateValue() many times


NhlanhlaHasane (approved):

codenamejason:
The hooks are now an npm package living in a different repo.

PR #290, opened 2021-07-28
pin antd version

azf20:
- ^ was using a version that left aligned menus


austintgriffith (approved):
thank you for finding this

PR #296, opened 2021-08-02
fix wallet.jsx

azf20:
Was using the provider to make transactions from the Account wallet, fixed to use the signer.


austintgriffith (approved):
🔥 🙏

Issue #245, opened 2021-07-02
Integrate command to auto-generate Interfaces from contracts


pabloruiz55:
As per: https://www.npmjs.com/package/hardhat-interface-generator
This package takes an specified contract and generates an Interface contract automatically.

Add the package and create a yarn command for convenience.

tabshaikh:
Giving this a shot :)

julianmrodri:
Implemented in #307

PR #307, opened 2021-08-04
add hardhat-interface-generator package

julianmrodri:
* Solves: #245 
* Adds the `hardhat-interface-generator` package that allows you to create interfaces based from a contract.
* Adds the required command so you can run `yarn gen-interface`
```bash
$ yarn gen-interface "YourContract"
```

```
yarn run v1.22.5
$ yarn workspace @scaffold-eth/hardhat gen-interface YourContract
$ hardhat gen-interface YourContract
Generated interface IYourContract
✨  Done in 1.61s.
```


packages/hardhat/package.json
```diff
@@ -38,6 +39,7 @@
     "balance": "hardhat balance",
     "send": "hardhat send",
     "generate": "hardhat generate",
+    "gen-interface": "hardhat gen-interface",
```
codenamejason, line 42: rename to "interface".

codenamejason (changes requested):
Everything looks good, I tested it on my local with no issues. Just need to rename it to "interface" and we are all good.

package.json
```diff
@@ -32,6 +32,7 @@
     "s3": "yarn workspace @scaffold-eth/react-app s3",
     "ship": "yarn workspace @scaffold-eth/react-app ship",
     "generate": "yarn workspace @scaffold-eth/hardhat generate",
+    "gen-interface": "yarn workspace @scaffold-eth/hardhat gen-interface",
```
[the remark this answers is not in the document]
julianmrodri: Done.

packages/hardhat/package.json
julianmrodri: Done

codenamejason (approved):

StErMi:
Yesterday @austintgriffith suggested to me this package and I was not totally sold by it. It's maybe fine to generate the interface of a "simple" contract but in my case where the contract was extending OpenZeppelin contracts, it generated a "faulty" interface that it was not usable. 

I had to delete it and write it manually by myself. I'll copy what I've said to @austintgriffith as a feedback

> I’ve tried it but it’s not good or at least not good for a complex contract like this that relies on external. it does not rely on OpenZeppelin interface and the code generated does not compile very well (a lot of warnings) because of the overloaded method, he does not add the SPDX-License and he does not add the pragma

PR #314, opened 2021-08-11
Upgrade for London

azf20:
Upgrade packages for new fee structure


codenamejason (commented):
./node_modules/ethers/lib.esm/ethers.js 20:0-22:22
Attempted import error: 'BaseContract' is not exported from '@ethersproject/contracts'.

When I update my project to the new version values I get the above error.

codenamejason (approved):
Looks good @azf20

PR #321, opened 2021-08-17
Challenge 1: Call out Hardhat bug fix

dmaulick:
Theres an issue with Hardhat where MetaMask mistakenly assumes all networks with http://localhost:8545 have a chain id of 1337
https://hardhat.org/metamask-issue.html

This made the challenge app crash on load.

For now, I am updating the code so that you can see exactly what needs to be done to resolve the issue. But I think this is probably better suited for a comment in the tutorial README


giorgosn:
This bug has burned my brain :P
are you planning adding it on the other branches?

codenamejason:
I think what everyone did for running local hardhat chain is add another network to MM using the 1337

PR #326, opened 2021-08-20
New network: Harmony ONE

givp:
## Change Details
Added Harmony testnet + mainnet networks. I have tested this end-to-end. Contract deployment and UI network switching function as expected. `yarn account` returns the correct balance.

## About Harmony
[Harmony](https://www.harmony.one) is a sharded and open blockchain. Harmony is 100% EVM compatible and runs Ethereum applications with 2-second transaction finality and 100 times lower fees.

## Network UI Preview
Image 2021-08-19 at 8 52 18 PM

### DIFF
--- a/packages/hardhat/hardhat.config.js
+++ b/packages/hardhat/hardhat.config.js
@@ -185,6 +185,22 @@ module.exports = {
         mnemonic: mnemonic(),
       },
     },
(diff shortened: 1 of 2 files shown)


austintgriffith (commented):
nice!!

PR #338, opened 2021-08-24
Fix scan / transfer display

carletex:
The address input is a bit unreachable. 

Before: 

image

After:
Screenshot from 2021-08-24 14-08-04

Thanks @dgrcode for the help.


owocki:
looks better. i think we might want to add a solid background to the div too.  its hard to read shit with the page background there

carletex:
@owocki Agreed. Tweaked the background color and some borders.

image

owocki:
👍

PR #353, opened 2021-08-30
Run prettier to fix lint errors

kenonchain:
Added a script to run prettier. So we can run `yarn prettier` to fix all the lint errors.

We could use https://www.npmjs.com/package/lint-staged to automate this process on commit hooks, but for an open-source project it might be opinionated.


VeryCB:
Cool! A third-party CI service can be integrated to check lint errors for every pull request as well.

beyondmtk:
code looks much better after `lint`.

codenamejason (approved):
I pulled this and ran `yarn prettier` and it ran successfully and I saw no issues with running the app locally.

PR #367, opened 2021-09-03
Update hardhat.config.js

macguyversmusic:
added example lines for web3 node entry via moralis.io so we are not limited to use infura.
https://rpc.maticvigil.com - if anyone is using this maybe they could also add the lines of config....


codenamejason (approved):
I like it!!

PR #372, opened 2021-09-03
fix buidl -> build in README.md

swekiiz:


codenamejason:
It is meant to be spelled that way. Buidl Guidl.

PR #390, opened 2021-09-08
issue # 389 - Fix: navigator.offLine returns false on surge hosted website

harryranakl:
navigator.offLine returns false on surge hosted website
considered an inconsistent API: Chrome and Safari it only determines whether a device has a connection to a network, not whether or not that network can actually reach the internet.


codenamejason:
@harryranakl is this an edge case for only surge? What if we deploy to fleek or netlify do we run into the same problem?

harryranakl:
> @harryranakl is this an edge case for only surge? What if we deploy to fleek or netlify do we run into the same problem?

effects browsers (like chrome and safari). not so dependent on hosting (like surge or netlify)

PR #392, opened 2021-09-08
Update issue templates

codenamejason:


viraj124 (approved):
LGTM @codenamejason :)

StErMi:
@codenamejason maybe we could add some more required info like

- scaffold-eth version (because they could not use the latest version)
- eth-hooks version
- <add all new direct scaffold-eth dependency> (like the new eth-components)

PR #396, opened 2021-09-10
Update a readme with a more long explanation

StErMi:
Deep dive into some concepts, probably the readme should be edited because this comes straight from my implementation and it has a lot of personal point of view (point of view from a newbie implementing it)


codenamejason (approved):
looks good, much better

PR #428, opened 2021-09-20
Added support to .env variables in both Hardhat and React app

StErMi:
Hi there everyone. I've configured both the Hardhat and React app package to use `.env` file.

You can see the example file in `/packages/hardhat/.example.env` and `/packages/react-app/.example.env`
When you fork the project you **MUST** change the name of those files to `.env` in order to let the system load them.

I have removed `InfuraId` from `Web3Modal` `walletconnect` because as far as I can see from their documentation (https://docs.walletconnect.org/quick-start/dapps/web3-provider#required) it's only required (the infuraid) if you are not specifying single RPC URLs. 
Because `scaffold-eth` is a general-purpose starter kit we should allow people to configure it as they need with Infura / Alchemy / Moralis.

I've tested only the basics but it should work.
Can someone more expert than me please test it correctly and try to deploy it on other networks configuring it?

On the Chrome console, I see some strange error logs but I don't think that the issue is because of these changes

1) POST https://rpc.scaffoldeth.io:48544/ net::ERR_CONNECTION_REFUSED -- Scaffold-eth RPC seems to be down?
2) 

```
ERROR LOADING CONTRACTS!! Error: processing response error (body="{\"jsonrpc\":\"2.0\",\"id\":42,\"error\":{\"code\":-32000,\"message\":\"the method has been deprecated: eth_accounts\"}}", error={"code":-32000}, requestBody="{\"method\":\"eth_accounts\",\"params\":[],\"id\":42,\"jsonrpc\":\"2.0\"}", requestMethod="POST", url="https://eth-mainnet.gateway.pokt.network/v1/lb/611156b4a585a20035148406", code=SERVER_ERROR, version=web/5.4.0)
    at Logger.makeError (index.ts:213)
    at Logger.throwError (index.ts:225)
    at index.ts:284
    at Generator.next (<anonymous>)
    at fulfilled (index.ts:2)
```

TODO

- [ ] check that everything works
- [x] try to deploy on testnet/other networks configuring the `.env` files
- [ ] double-check those 2 errors if they are related to my changes


StErMi:
@codenamejason could you do the code check?
@captnseagraves we should update the Documentation site adding the `.env` part where we explain that they should change the name of those files and configure them?

codenamejason:
@StErMi I will do the check today :)

codenamejason:
Screen Shot 2021-09-20 at 10 18 00 AM

yarn deploy is failing @StErMi

codenamejason:
Okay, it was just a typo..

codenamejason (approved):
ran local and deployed and interacted with contracts on localhost and kovan.

codenamejason:
I am going to wait for the docs to be updated before I merge this @StErMi @captnseagraves

StErMi:
Hi there, before the merge we need to think carefully on how we should notify the user that he needs to change the name of the `.example.env` from both the `react-app` and `hardhat` package to `.env` and he needs to update the needed params in order to deploy/use a specific chain.

What would be the best way to do that? 

Another change that I've made (and tell me if it was correct or not) was this: previously we had

```js
const localProviderUrl = targetNetwork.rpcUrl;
const localProviderUrlFromEnv = process.env.REACT_APP_PROVIDER ? process.env.REACT_APP_PROVIDER : localProviderUrl;
const localProvider = new ethers.providers.StaticJsonRpcProvider(localProviderUrlFromEnv);
```

I switched to only use the `targetNetwork.rpcUrl`

```js
const localProviderUrl = targetNetwork.rpcUrl;
const localProvider = new ethers.providers.StaticJsonRpcProvider(localProviderUrl);
```

In theory, the user needs to configure the `.env` that is used by the `targetNetwork` and we don't need the generic `process.env.REACT_APP_PROVIDER`

Let me know what you think

codenamejason:
@austintgriffith what are your thoughts on this? I know it will be more secure and have everything in one place, the only side effects I see should be with the documentation and walking the user through the new setup process using env vars.

codenamejason:
@StErMi read the comment above the rpc setups, we need to still make sure everything works without using the env as well. 

This is set up wrong. I added defaults instead of null in this image but we need to try and use the Scaffold-Eth provider first and then the fallback if needed.

Screen Shot 2021-09-27 at 9 17 47 AM

StErMi:
> @StErMi read the comment above the rpc setups, we need to still make sure everything works without using the env as well.
> 
> This is set up wrong. I added defaults instead of null in this image but we need to try and use the Scaffold-Eth provider first and then the fallback if needed.
> 
> Screen Shot 2021-09-27 at 9 17 47 AM

yeah totally correct I'll merge the new scaffold-eth branch and manage defaults like you have done

PR #429, opened 2021-09-20
App.jsx changes to work with staking contract example

brendanboyle87:
Ethers.utils and duration imported fixed some undefined errors. 
Removed address from withdraw function as example uses msg.sender.


codenamejason:
looks good

PR #445, opened 2021-09-24
hardhat node command typo

3scava1i3r:
changed command


codenamejason:
Should be using yarn chain for the local chain. Why are you using yarn node? I cannot merge this, not the way to do it.

3scava1i3r:
@codenamejason why is there a node command to begin with then? If we can have local chains using 'chain' command and fork using fork command?

PR #455, opened 2021-09-27
Contract component function overloading fix

azf20:
- Fix for overloaded functions in the Contract component -> no overloaded functions would show up as function lookup on the contract object was undefined (needed to be the full signature with args)

- e.g. loogie-tank branch on Mainnet LOOG tab


codenamejason (approved):

codenamejason:
Thanks @azf20

PR #473, opened 2021-10-07
Fix README Typo

Hakeemmidan:
(Not sure if this is a typo or not.)

Changed 'buidl' to 'build'.


captnseagraves:
Thanks @Hakeemmidan. It is not a typo. Scaffold-eth is built and maintained by the Buidl Guidl. Like HODL.

PR #481, opened 2021-10-10
Add jazzicon component

danfinlay:
Uses same interface as blockie component.

The thing is, a number of existing components are currently hard-coded for the Blockie format:

- Address
- AddressInput
- Contract/FunctionForm
- Faucet
- Timeline
- index

Not sure what migration path you'd prefer for allowing identicon customization. One low-obligation way would be to allow passing a `identicon={MyComponent}` parameter to all of those components, so that a parent could always push in a preferred identicon format.

Another approach might be to allow them all to tap into some global context.

For now, just opening this because it does the first step, and should be safe to merge, but would be best paired with an answer to that question.

Simple usage today: Replace the `Blockie.jsx` file with the `Jazzicon.jsx` file, and your project will use jazzicons instead.

### DIFF
--- a/packages/react-app/src/components/Jazzicon.jsx
+++ b/packages/react-app/src/components/Jazzicon.jsx
@@ -2,7 +2,7 @@ import React from "react";
 import Jazzicon from "react-jazzicon";
 import { toChecksumAddress } from "ethereumjs-util";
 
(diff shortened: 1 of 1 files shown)


danfinlay:
Closing for now, there are some tweaks this needs.

PR #483, opened 2021-10-10
Subgraph: rollback graph-node, graph-cli and api version to stable versions (Fix for Mac M1 systems)

0xGabi:
This PR rollback to a stable version where the subgraph service functionality is working.

Changes introduced in https://github.com/scaffold-eth/scaffold-eth/pull/437 and https://github.com/scaffold-eth/scaffold-eth/pull/457/files are not working. The local graph-node service is not indexing events and throwing with:
```
failed to process trigger: ... Mapping terminated before handling trigger: oneshot canceled
```

### DIFF
--- a/packages/services/graph-node/docker-compose.yml
+++ b/packages/services/graph-node/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "3"
 services:
   graph-node:
-    image: graphprotocol/graph-node:latest
+    image: graphprotocol/graph-node:2c23cce
     ports:
       - "8000:8000"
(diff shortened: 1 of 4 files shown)


codenamejason:
Is there a way to get it working using the newer version? Using the older version was causing other dependency issues on yarn install (causing it to fail). @azf20 I would love your graph skills here ser and get your thoughts.

0xGabi:
@azf20 I didn't saw that issue. You are right, I am using a Mac M1 and experienced that exact error. Closing this one!

PR #485, opened 2021-10-11
Adding Docker setup

arturgontijo:
This PR adds a Docker alternative setup to spin up the stack inside a container.

If it makes sense to have, I'll need guidance about where to add current `docker/README.md` content.


arturgontijo:
> It's not leveraging the `auto reload` logic, fixing it.

Done.

PR #488, opened 2021-10-11
chore: sync yarn.lock

meetmangukiya:
Running `yarn` on master yields some uncommitted changes to `yarn.lock`


codenamejason:
yarn.lock should not even be included in the repo, this file should exist to you locally.

codenamejason:
I know, it keeps getting back in there... it is supposed to exist locally though. I just revert changes on yarn.lock before I check in my changes so it is not includes seeing how adding to .gitignore is not always working...

PR #525, opened 2021-10-30
Example for testing events

moltam89:
Hey guys,

It took me a while to figure it out how this can be done, maybe someone else will find this useful.
I commented out the new code, because it only makes sense when the SetPurpose event is emitted from YourContract.

Best,
Tamas


codenamejason (approved):
great addition for the new dev!!

PR #529, opened 2021-11-02
fix: remove OVM configuration from kovanOptimism

smartcontracts:
Optimistic Kovan now runs pure EVM so the OVM configuration is no longer required (will actually break if enabled). Gas price can also no longer be zero.

### DIFF
--- a/packages/hardhat/hardhat.config.js
+++ b/packages/hardhat/hardhat.config.js
@@ -206,11 +206,9 @@ module.exports = {
     },
     kovanOptimism: {
       url: "https://kovan.optimism.io",
-      gasPrice: 0,
       accounts: {
         mnemonic: mnemonic(),
       },
-      ovm: true,
       companionNetworks: {
         l1: "kovan",
       },


smartcontracts:
Screenshot of `scaffold-eth` working after making this change:
Screenshot from 2021-11-02 12-51-34

azf20:
Aha there is me reading my inbox chronologically, hence missing this when I made [that comment](https://github.com/scaffold-eth/scaffold-eth/issues/528#issuecomment-958051552) ! 👁️ now...

azf20 (approved):
Looks good! Presumably when mainnet is upgraded, we can do the same for optimism mainnet & remove the @eth-optimism/hardhat-ovm dependency?

smartcontracts:
> Looks good! Presumably when mainnet is upgraded, we can do the same for optimism mainnet & remove the @eth-optimism/hardhat-ovm dependency?

Yep! Exactly. I'll make sure to create a PR as soon as mainnet is upgraded :-)

PR #543, opened 2021-11-09
feat: add UI handlers and fix contract logic

ironsoul0:


viraj124 (changes requested):
hey, can we also display the current floor on UI if not done till now

PR #545, opened 2021-11-10
fix post master merge

brendanboyle87:


codenamejason:
Screen Shot 2021-11-09 at 8 37 46 PM

you only need one of the checks

PR #549, opened 2021-11-11
Challenge 2 token vendor typo in instructions

TaeHyunJin:
Reduces confusion


codenamejason (approved):

codenamejason:
we are moving the challenges to scaffold-eth/scaffold-eth-challenges, it is a separate repo and would be grateful if you were to update that one as well @TaeHyunJin

PR #553, opened 2021-11-11
Challenge 3 dex fix react dex component

mysliwiec-tech:
While going by step by step instructions, I have found some bugs or missing code that I think might make it difficult for a user to finish challenge if that person is not familiar with react (_Checkpoint 6: UI_ code is broken).

`dex` variable was missing in the `00_deploy_your_contract.js` script + there was a lot of issues that I addressed, that were related to some imports missing or wrong order of parameters in the react app (most of it in `DEX.jsx`) + injectedProvider has not been passed to the component. Please review attached commits. Tested on my environment and it works. Maybe not the most optimal way, but I'm not an expert.

Still there is an error being thrown by _safeMath_ lib from `DEX.sol` when withdrawing the liquidity from dex, but that will be addressed in another pull request next time. Also chart does not seem to work.

### DIFF
--- a/packages/hardhat/deploy/00_deploy_your_contract.js
+++ b/packages/hardhat/deploy/00_deploy_your_contract.js
@@ -21,6 +21,7 @@ module.exports = async ({ getNamedAccounts, deployments }) => {
     log: true,
   });
 
+  const dex = await ethers.getContract("DEX", deployer);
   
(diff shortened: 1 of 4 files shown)


codenamejason:
I have merged but we are moving to this repo to the challenges repo.  scaffold-eth/scaffold-eth-challeges

PR #556, opened 2021-11-12
add changeNetwork.py

leeduckgo:
change network of hardhat & react-app by python script.


codenamejason:
we cannot merge this with the current information. What is the purpose? Why Python for a javascript project?

PR #566, opened 2021-11-15
Fix optional render of connect button

ghostffcode:
Fix issue with seeing "Connect" button on other components using the `Account.jsx` component

Screen Shot 2021-11-14 at 10 56 49 PM


codenamejason (approved):
Thanks Bliss!

PR #576, opened 2021-11-18
App.jsx clean up

ghostffcode:
- Extracts `faucetHint` & `NetworkDisplay` logic and UI into components
- Create web3Modal setup into helpers
- Add `Home.jsx` as the default view when app loads


codenamejason (approved):
looking good 🥷🏽

PR #581, opened 2021-11-26
fix rendering of plot

andrzejKraw:


codenamejason (dismissed):

codenamejason:
we are now maintaining the challenges on the scaffold-eth-challenges repo. Would you mind pr'ing to that repo? @andrzejKraw the dex challenge is now challenge 5

codenamejason:
wrong repo but will update for the users still using. This will take time to move everyone to the new repo.

andrzejKraw:
Thanks, I wasn't aware of the new repo. 
I see that this fix is not needed there though.

PR #582, opened 2021-11-26
Punk Wallet Network Fix

jfarid27:
# Description

Punk wallet is currently broken for transactions due to the replay-protection update. It appears that replay protected transactions require the chainID when forming the transaction. Can someone confirm if this works? You should be able to send Matic via the punk-wallet app.


codenamejason:
this branch will now be maintained in the scaffold-eth-examples repo @jfarid27 would you mind pr'ing to that one?

codenamejason (approved):
thank you

PR #591, opened 2021-12-03
allow for readable display of address value when it is a response to …

dvinubius:
…a view/pure function call

The FunctionForm displays its result with tryToDisplay(). 

In case of pure/view functions where the response is an address, tryToDisplay() creates an Address component.

Since that is not readable in the FunctionForm, a special handling to display this address value as text can improve the UX.


dvinubius:
ISSUE:
Screenshot 2021-12-03 at 22 00 31
)

codenamejason (approved):

PR #601, opened 2021-12-07
Network UX

dvinubius:
**Improve network switch initialization**
It selects your intended network instead of defaulting to the first option and alerting you in case it doesn't match.

**Improve network mismatch alert** -
If you want to work on localhost but are logged in with an injected provider the alert suggests you switch to localhost. I found some things don't work as expected when working locally via metamask (like switching accounts and having correct balance updates).
So I've added a special handling for this case: It suggest you logout from the injected provider.


codenamejason (approved):
🔥

PR #602, opened 2021-12-07
Add Optimism kovan and mainnet chain ids to transactor notify

damianmarti:


codenamejason (approved):
Thanks brother 🔥

damianmarti:
@codenamejason I think that we should revert this PR. Notify doesn't support Optimism yet, so the callback set in the transaction doesn't called on every status change.

codenamejason:
@damianmarti is is causing errors now for anyone? I can revert if they are.

damianmarti:
@codenamejason Only if you are using Optimism network. If you add a callback on a Transactor call, this will not work as you expected since the callback will be called just one time and not one on each status change.

PR #610, opened 2021-12-15
Hooks renaming

Munsterberg:
I'm sure these aren't widely used, but it was bothering me. I use a fuzzy finder and searching `useX` is just more intuitive for me, hopefully, others agree. This simplifies things a bit when quickly searching for the hook to make changes or to inspect the functionality. 

I do this when I scaffold a new project anyways, but figured why not open a PR and see if others would find this as a small QoL improvement.


codenamejason:
I agree with the consistency, thanks for the PR!

PR #614, opened 2021-12-17
Add Truffle

fainashalts:
This PR adds Truffle to the scaffold-eth project, replacing scripts that were previously using HH. Scripts can be found in `packages/truffle/scripts`. 

Work remains on the react side of this project, as we need to decide how to approximate the single file that contains HH built contracts should be generated by Truffle. However, as it stands all scripts in the package.json should work.


Silur:
how come this was never merged?

PR #620, opened 2021-12-22
Revert "Add Optimism kovan and mainnet chain ids to transactor notify"

codenamejason:
Reverts scaffold-eth/scaffold-eth#602


codenamejason:
If you add a callback on a Transactor call, this will not work as you expected since the callback will be called just one time and not one on each status change.

PR #625, opened 2021-12-30
fix block explorer in contract component

artdgn:
This fixes using the correct block explorer in Contract component (address views).

On the way it also fixes Harmony name in constants to be as the object key (app depends on it to get the RPC correctly).


codenamejason:
dope thanks

codenamejason (approved):

PR #631, opened 2022-01-06
Fix: No errors in the console

carletex:
A while ago, we removed the "airbnb" linting in [our eslint config file](https://github.com/scaffold-eth/scaffold-eth/blob/master/packages/react-app/.eslintrc.js#L6-L7). I think it was removed because the code wasn't compiling when encountering linting errors (which should be marked as warnings... not errors).

The issue with this approach is that now **we are not getting real errors feedback** in the console: syntax errors, forgotten imports, using undeclared variables, etc. We only get prettier (styling) warnings.

e.g: If I remove the `useState` import.

In the console I get:

image

When I go to the browser:

image

I think the feedback on the console is super helpful and we should bring that back.

## Proposed solution

CreateReactApp provides the ESLint packages + a set of rules (less opinionated that the airbnb ones) out of the box. You can extend and customize them (check `eslintConfig` inside react-app/package.json). Some docs here: https://create-react-app.dev/docs/setting-up-your-editor/#extending-or-replacing-the-default-eslint-config

So in this PR:

- [x] Removed unnecessary eslint packages
- [x] Remove custom `.eslintrc.js` file
- [x] Enable CRA eslint + prettier. 

Now, for the example above, we get:
image

--- 

Now, when no errors, we are getting **all** the linting **warnings** (but the code compiles, as it should be). **I was thinking on amending all these warnings in a separate PR if/when this one gets merged.**


codenamejason (approved):
Dude, thank you so much 🔥

codenamejason:
Tested locally and speedran it with success.

PR #636, opened 2022-01-11
Prevent page from moving when scrollbar appears / disappears

RogerPodacter:
For example, when the wallet modal appears the page height is 100% and so there is no scrollbar. But then when the modal disappears the scrollbar comes back and the page shifts. This CSS is a transparent solution to the problem (see https://stackoverflow.com/a/30293718/25068)


codenamejason (approved):
awesome, thanks for this 🔥

PR #648, opened 2022-01-18
Add support for Fantom mainnet and testnet

Warkanlock:
Since Fantom is entirely EVM compatible it would be nice to include the network in the configuration files of ``hardhat``.

I added Fantom configurations in order to use ``testnet`` or ``mainnet`` for Fantom. 

In order to use this config, just set up inside ``hardhat.config.js`` the defaultNetwork to point out to Fantom

image

You can set: **fantom** for ``mainnet`` and **testnetFantom** for ``tests``

It's likely that I would create a tutorial to use this network within **scaffold-eth**

If you don't know Fantom, you can read more here: https://fantom.foundation/

### DIFF
--- a/packages/hardhat/hardhat.config.js
+++ b/packages/hardhat/hardhat.config.js
@@ -152,6 +152,20 @@ module.exports = {
         mnemonic: mnemonic(),
       },
     },
+    fantom: {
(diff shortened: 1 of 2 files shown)


codenamejason (approved):
Thank you ser 🍺

PR #665, opened 2022-02-02
Fixed token selector duplicating items

### LINKED ISSUE #632: Token List is acting funky
**Describe the bug**

Screen Shot 2022-01-06 at 7 05 26 PM

**To Reproduce**

- [ ] yarn chain
- [ ] yarn deploy
- [ ] yarn start
- [ ] navigate to 'Hints' tab
- [ ] use the Pick a token select list and scroll

MimmDev:
fixes #632 

The issue was caused by `listOfTokens` containing tokens which have duplicate symbols (CMI, for example). As these symbols were not unique but used as keys for the `Option` components, it caused React to duplicate/overwrite list items while scrolling.

### DIFF
--- a/packages/react-app/src/views/Hints.jsx
+++ b/packages/react-app/src/views/Hints.jsx
@@ -113,7 +113,7 @@ export default function Hints({ yourLocalBalance, mainnetProvider, price, addres
           optionFilterProp="children"
         >
           {listOfTokens.map(token => (
-            <Option key={token.symbol} value={token.symbol}>
+            <Option key={token.address} value={token.symbol}>
               {token.symbol}
             </Option>
           ))}


packages/react-app/src/views/Hints.jsx
```diff
@@ -113,7 +113,7 @@ export default function Hints({ yourLocalBalance, mainnetProvider, price, addres
           optionFilterProp="children"
         >
           {listOfTokens.map(token => (
-            <Option key={token.symbol} value={token.symbol}>
+            <Option key={token.address} value={token.symbol}>
```
kamikazebr, line 116: Nice catch! Idk if is to much, but you could even concatenate symbol more address there to even avoid address conflict. (i know looks impossible that to happend). Up to you.

kamikazebr (approved):
Nice catch!

packages/react-app/src/views/Hints.jsx
codenamejason: I updated to be key={token.address + "_" + token.symbol}

codenamejason (approved):

PR #694, opened 2022-02-20
tokenlist minor bug

zikyfranky:
I noticed whenever a chain id is passed to the useTokenList hook, the returned list is always undefined due to the fact that it's being set to `_tokenList.tokens` after filtering by `chainId`.


Jcauley (commented):
looking good i think i need to redo something now with my end

Jcauley:
the difficulty of blocks to come out for the pull space if that makes any different with the hash rate if the last 14day period

PR #706, opened 2022-03-10
PunkWallet doesn't work on Polygon

moltam89:
Hi @austintgriffith ,

I noticed that I cannot transfer funds with PunkWallet on the Polygon chain.
After debugging the code it turned out to be an ethers bug, which is now fixed in the latest version - 5.5.6.

Here is the ethers issue:
https://github.com/ethers-io/ethers.js/issues/2691

**Note:**
Actually PunkWallet still doesn't work on Polygon, because the gasPrice is set to 1gwei, which is not enough right now.
I can look into this later, manually setting the gasPrice worked for me, my transaction was confirmed.

Best,
Tamas


Jacquelinevv0693:
Duplicate of #2691

PR #711, opened 2022-03-16
make Header component more dynamic

pankaj512:
instead of hard coding the link, title and
subtitle, it is replace with default props. This
makes header component more dyanmic


codenamejason (commented):
I like this, let me test and I will merge

codenamejason (approved):

PR #725, opened 2022-03-27
pin ipfs version, add note on chain running

azf20:
Currently breaks on startup using more recent ipfs version, this pins the version and adds an additional hint to the subgraph page


codenamejason (approved):

codenamejason:
Thanks @azf20

PR #749, opened 2022-04-21
Create react-app/src/contracts directory if it does not already exist

nullbitx8:
This fixes the script so it doesn't blow up on the first run


codenamejason:
The contracts directory will always exist with the external_contracts.js file in it @nullbitx8

nullbitx8:
@codenamejason perhaps I had something messed up in my local environment when I opened this. Thanks!

PR #751, opened 2022-04-23
Add hardhat-abi-exporter plugin

pankaj512:
Added hardhat-abi-exporter to dependency list of hardhat. This helps when there are multiple contracts and one contract is uses as state inside another. At deploy time we can only export deployed contracts abi but sometimes need arises where you want to call function of contract that is used as state inside deployed contract. 

This plugin exports abi of all the contracts inside "../react-app/src/contracts" with same name as contracts inside a json file for each contract. You can just export abi and create the Read/WriteContract using the address of state contacts and abi.

Usecase:
MultiISigWallet uses as state inside MultiSigFactory and MultiSigFactory gets deployed and you dynamically create MultiISigWallet and want to call function of MultiISigWallet

### DIFF
--- a/packages/hardhat/hardhat.config.js
+++ b/packages/hardhat/hardhat.config.js
@@ -8,6 +8,7 @@ require("@tenderly/hardhat-tenderly");
 
 require("hardhat-deploy");
 require("hardhat-gas-reporter");
+require("hardhat-abi-exporter");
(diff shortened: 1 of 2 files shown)


ghostffcode (approved):
Thanks for adding this

PR #762, opened 2022-04-29
Revert "Bump url-parse from 1.5.3 to 1.5.10"

inosgun:


codenamejason:
@inosgun why do you want to revert this?

PR #765, opened 2022-04-30
Require msg.value >= 0.01 && msg.value <= 0.01

chrontast:


codenamejason:
wouldn't you want it to be:
``` soldiity
require(msg.value == 0.01 ether, "Please send exactly 0.01 ETH. Thanks!");
```

PR #768, opened 2022-05-04
bump graph & tidy docker

azf20:
- update graph-cli and graph-ts to the latest version
- use the built-in network config rather than mustache


austintgriffith (approved):
👍

PR #792, opened 2022-05-27
Make the Contract component work with tuples

azf20:


codenamejason:
dope

codenamejason (approved):

chris13524:
Seems like this PR may have caused a bug? See #800

PR #819, opened 2022-07-06
Faucet.jsx - Allow multiple calls

ZakGriffith:


bhndt (approved):
0xc6B499931c1E12Ab047EF5cb98963040B002421B

codenamejason:
@ZakGriffith we could also increase the amount to distribute from the faucet at one time.

PR #821, opened 2022-07-10
Added support for binance smart chain

KlimMixer:
This PR fixed version to support of binance smart chain, with hardhat configs.


bhndt (approved):
master

PR #825, opened 2022-07-14
feat: add optional scaffold-eth-cli command

QEDK:
This PR is intended as a temperature test for:
* providing traction to `scaffold-eth-cli` to further determine,
* if using `scaffold-eth-cli` is helpful for repeat iterators
  * think instead of having to find the right thing to clone everytime, you could simply do: `npx scaffold-eth nextjs-typescript` or `npx scaffold-eth <branch-name>`, maybe with some fuzzy searching algo
* if `scaffold-eth-cli` can be absorbed into the `scaffold-eth` org as a product to take forward 

cc @codenamejason @austintgriffith

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,12 @@ Prerequisites: [Node (v16 LTS)](https://nodejs.org/en/download/) plus [Yarn (v1.
 git clone https://github.com/scaffold-eth/scaffold-eth.git
 ```
 
+or using `scaffold-eth-cli`:
+
+```bash
+npx scaffold-eth
+```
+
 > install and start your 👷‍ Hardhat chain:
 
 ```bash


ghostffcode:
Is the repo for the CLI public?

QEDK:
> Is the repo for the CLI public?

yep, see https://github.com/QEDK/scaffold-eth-cli

codenamejason (dismissed):

codenamejason:
I can make a repo under Scaffold-Eth for the cli code to live @QEDK wdyt?

QEDK:
> I can make a repo under Scaffold-Eth for the cli code to live @QEDK wdyt?

works for me, I can transfer the repo to the org, but I would still like to remain as the code owner and maintainer, if that's okay.

PR #843, opened 2022-08-05
add Jest settings

skaunov:
these settings let run Jest tests for React-app


skaunov:
@codenamejason , I don't see anything to do further here aside of review and merging, and these are I don't have privileges to. X)

PR #844, opened 2022-08-05
delete file App.test.js from React-app

skaunov:
Delete testing template (added by create-react-app) which isn't designed to work with Scaffold-eth.

This template is more misleading than helpful atm. A better solution would be to adapt it to Scaffold. If it's needed, please reply in discussion what kind of adaptation would be merged, so that PR could be improved.

### DIFF
--- a/packages/react-app/src/App.test.js
+++ b/packages/react-app/src/App.test.js
@@ -1,9 +1,0 @@
-import React from "react";
-import { render } from "@testing-library/react";
-import App from "./App";
-
-test("renders learn react link", () => {
-  const { getByText } = render(<App />);
-  const linkElement = getByText(/learn react/i);
-  expect(linkElement).toBeInTheDocument();
-});


codenamejason (approved):
I agree, I don't think anyone ever uses the test file for the font-end. it's something we never really set up. I would be happy to chat about a testing framework that would work with Scaffold-Eth 🏗

skaunov:
@codenamejason , I ran React Testing Library on Scaffold-Eth :building_construction: while doing MultiSig challenge (saved the basic settings in #843). Though I only managed to reach correct response from Jest that RTL is doing its job. For actual tests to work it would lead to (a lot of) refactoring.

By adaptation I meant something as simple as CRA test template itself. "Hello world" variant of test-case which works in Scaffold-Eth :building_construction:.

And creating a testing suite for Scaffold doesn't look as a small task. I would tackle this one from assessing what parts (probably components) are most ready to be covered by tests (i.e. with minimum of refactoring), and which are most critical (if any) to be focused with testing (including "snapshots" = regression): to have something to start with. Just off the top of my head it could be e2e CRA tests for general user tasks (wallet, sending eth, and so on), and jest/enzyme tests for components that are included with Scaffold-Eth. Then it seems to be a good practice to support _snapshot_ test, which are the way React is tested for regressions. Also there are some tools for hooks testing, but I don't think this will go all the way down to hooks testing. =)

PR #851, opened 2022-08-14
Import functionality from Punk Wallet

moltam89:
This PR adds the import functionality, code is from Punk Wallet, https://github.com/scaffold-eth/punk-wallet/pull/11

image


moltam89:
I pushed extra commits to remove punks as it was requested.
Also I replaced the old xdai.io link with Punk Wallet.

PR #869, opened 2022-09-03
Revert "Update Hardhat"

codenamejason:
Reverts scaffold-eth/scaffold-eth#866


codenamejason:
Screen Shot 2022-09-03 at 7 51 51 PM

was having this issue after merge. Can you double check @azf20 if this is happening to you?

PR #873, opened 2022-09-06
Remove eth-optimism sdk

azf20:
Was causing `node-gyp` errors on `yarn install`


codenamejason (approved):
LGTM, and if someone needs it down the road they can add it.

PR #881, opened 2022-09-13
Prettier errors and button fix

ZakGriffith:
Cleaned prettier errors in header.jsx, and updated the Roll The Dice button to enable every time.

### DIFF
--- a/packages/react-app/src/App.jsx
+++ b/packages/react-app/src/App.jsx
@@ -439,11 +439,10 @@ function App(props) {
   useOnBlock(localProvider, () => {
     if (DEBUG) console.log(`⛓ A new local block is here: ${localProvider._lastBlockNumber}`);
     if (DEBUG) console.log("blockNumber: ", blockNumber);
-    setRollDisabled(
-      blockNumber === 0 ||
-        ethers.BigNumber.from(localProvider._lastBlockNumber) < blockNumber.add(futureBlocks) ||
-        ethers.BigNumber.from(localProvider._lastBlockNumber) > blockNumber.add(futureBlocks + 256),
-    );
+    const lastBlock = parseInt(ethers.BigNumber.from(localProvider._lastBlockNumber).toString());
+    const betBlock = parseInt(blockNumber.toString());
+    const disable = betBlock === 0 || lastBlock < betBlock + futureBlocks || lastBlock > betBlock + futureBlocks + 256;
+    setRollDisabled(disable);
   });
 
   const bet = async () => {
--- a/packages/react-app/src/components/Header.jsx
+++ b/packages/react-app/src/components/Header.jsx
@@ -5,12 +5,12 @@ import React from "react";
 
 export default function Header() {
   return (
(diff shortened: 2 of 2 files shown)


packages/react-app/src/App.jsx
```diff
@@ -439,11 +439,10 @@ function App(props) {
   useOnBlock(localProvider, () => {
     if (DEBUG) console.log(`⛓ A new local block is here: ${localProvider._lastBlockNumber}`);
     if (DEBUG) console.log("blockNumber: ", blockNumber);
-    setRollDisabled(
-      blockNumber === 0 ||
-        ethers.BigNumber.from(localProvider._lastBlockNumber) < blockNumber.add(futureBlocks) ||
-        ethers.BigNumber.from(localProvider._lastBlockNumber) > blockNumber.add(futureBlocks + 256),
-    );
+    const lastBlock = parseInt(ethers.BigNumber.from(localProvider._lastBlockNumber).toString());
```
damianmarti, line 442: Maybe is better to compare BigNumbers to avoid errors if the blockNumber increase more than JavaScript integer can manage.

ZakGriffith: That is what I thought as well, but blockNumber.add(futureBlocks) wasn't returning anything so half the time the button stayed disabled when the correct block was mined.  Is it working for you?

damianmarti: We should change the compare to use something like 

ethers.BigNumber.from(localProvider._lastBlockNumber).lt(blockNumber.add(futureBlocks))

ZakGriffith: Been adding some output.  this line:
console.log("blockNumber.add(futureBlocks)", blockNumber.add(futureBlocks));

Throws a front end error of:
TypeError: blockNumber.add is not a function

Note sure why, but it may have something to do with why the button stays disabled.

damianmarti: Maybe blockNumber is not loaded yet. This is why I was checking "blockNumber === 0 ||"

ZakGriffith: blockNumber is there, I logged that first just top make sure.  I thought blockNumber === 0 was there to make sure there was an actual betting block going on.  That seems to get reset to 0 after each betting round.

damianmarti: Yes, after the dice is rolled, blockNumber is set to 0.

damianmarti: Can you please try it?

    setRollDisabled(
      blockNumber === 0 ||
        ethers.BigNumber.from(localProvider._lastBlockNumber).lt(blockNumber.add(futureBlocks)) ||
        ethers.BigNumber.from(localProvider._lastBlockNumber).gt(blockNumber.add(futureBlocks + 256)),
    );

damianmarti: It seem it's working ok at https://dice-random.surge.sh/

ZakGriffith: Yeah, that 100% fixed it.  How?  Nice work.

PR #883, opened 2022-09-15
Hotfix/  rpc polling issue proposal 1

Naim-Bijapure:
#CHANGES MADE FOR THIS ISSUE
https://github.com/scaffold-eth/scaffold-eth/issues/882
changes: 
1. In this pr changed useStaticJsonRPC hook. 
instead of initiating 2-times useStaticJsonRPC hook for localProvider and mainnetProvider  accepting 
localUrl, urlArray, isMainnet as parameters and giving back localProvider and mainnetProvider at 1 call. 
2. also added logic for if it is mainnetHook then  use localProvider as mainnetProvider only creating 1 provider. 
3. added loadMainnetProvider function inside useStaticJsonRPC to get the mainnetProvder from an array. at old logic, the promise.race()  calls all the rpc's in array . but in loadMainnetProvider it calls first RPC if it detects the network and then skips other 2 calls else it will call another RPC in array etc. 
4. other minor changes like polling time etc done. 

2022_September_15_18-29-15

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,3 +120,187 @@ Join the telegram [support chat 💬](https://t.me/joinchat/KByvmRe5wkR-8F_zz6Aj
 ### Automated with Gitpod
 
 [![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/#github.com/scaffold-eth/scaffold-eth)
+
+
+
+      
+
+      {/* ✏️ Edit the header and change the title to your project name */}
+      <Header>
+        {/* 👨‍💼 Your account is in the top right with a wallet at connect options */}
+        <div style={{ position: "relative", display: "flex", flexDirection: "column" }}>
+          <div style={{ display: "flex", flex: 1 }}>
+            {USE_NETWORK_SELECTOR && (
+              <div style={{ marginRight: 20 }}>
+                <NetworkSwitch
+                  networkOptions={networkOptions}
+                  selectedNetwork={selectedNetwork}
+                  setSelectedNetwork={setSelectedNetwork}
+                />
+              </div>
+            )}
+            <Account
+              useBurner={USE_BURNER_WALLET}
+              address={address}
+              localProvider={localProvider}
+              userSigner={userSigner}
+              mainnetProvider={mainnetProvider}
+              price={price}
+              web3Modal={web3Modal}
+              loadWeb3Modal={loadWeb3Modal}
+              logoutOfWeb3Modal={logoutOfWeb3Modal}
+              blockExplorer={blockExplorer}
+            />
+          </div>
+        </div>
+      </Header>
+      {yourLocalBalance.lte(ethers.BigNumber.from("0")) && (
+        <FaucetHint localProvider={localProvider} targetNetwork={targetNetwork} address={address} />
+      )}
+      <NetworkDisplay
+        NETWORKCHECK={NETWORKCHECK}
+        localChainId={localChainId}
+        selectedChainId={selectedChainId}
+        targetNetwork={targetNetwork}
+        logoutOfWeb3Modal={logoutOfWeb3Modal}
+        USE_NETWORK_SELECTOR={USE_NETWORK_SELECTOR}
+      />
+      <Menu style={{ textAlign: "center", marginTop: 20 }} selectedKeys={[location.pathname]} mode="horizontal">
+        <Menu.Item key="/">
+          <Link to="/">App Home</Link>
+        </Menu.Item>
+        <Menu.Item key="/debug">
+          <Link to="/debug">Debug Contracts</Link>
+        </Menu.Item>
+        <Menu.Item key="/hints">
+          <Link to="/hints">Hints</Link>
+        </Menu.Item>
+        <Menu.Item key="/exampleui">
+          <Link to="/exampleui">ExampleUI</Link>
+        </Menu.Item>
+        <Menu.Item key="/mainnetdai">
+          <Link to="/mainnetdai">Mainnet DAI</Link>
+        </Menu.Item>
+        <Menu.Item key="/subgraph">
+          <Link to="/subgraph">Subgraph</Link>
+        </Menu.Item>
+      </Menu>
+
+      <Switch>
+        <Route exact path="/">
+          {/* pass in any web3 props to this Home component. For example, yourLocalBalance */}
+          <Home yourLocalBalance={yourLocalBalance} readContracts={readContracts} />
+        </Route>
+        <Route exact path="/debug">
+          {/*
+                🎛 this scaffolding is full of commonly used components
+                this <Contract/> component will automatically parse your ABI
+                and give you a form to interact with it locally
+            */}
+
+          <Contract
+            name="YourContract"
+            price={price}
+            signer={userSigner}
+            provider={localProvider}
+            address={address}
+            blockExplorer={blockExplorer}
+            contractConfig={contractConfig}
+          />
+        </Route>
+        <Route path="/hints">
+          <Hints
+            address={address}
+            yourLocalBalance={yourLocalBalance}
+            mainnetProvider={mainnetProvider}
+            price={price}
+          />
+        </Route>
+        <Route path="/exampleui">
+          <ExampleUI
+            address={address}
+            userSigner={userSigner}
+            mainnetProvider={mainnetProvider}
+            localProvider={localProvider}
+            yourLocalBalance={yourLocalBalance}
+            price={price}
+            tx={tx}
+            writeContracts={writeContracts}
+            readContracts={readContracts}
+            purpose={purpose}
+          />
+        </Route>
+        <Route path="/mainnetdai">
+          <Contract
+            name="DAI"
+            customContract={mainnetContracts && mainnetContracts.contracts && mainnetContracts.contracts.DAI}
+            signer={userSigner}
+            provider={mainnetProvider}
+            address={address}
+            blockExplorer="https://etherscan.io/"
+            contractConfig={contractConfig}
+            chainId={1}
+          />
+          {/*
+            <Contract
+              name="UNI"
+              customContract={mainnetContracts && mainnetContracts.contracts && mainnetContracts.contracts.UNI}
+              signer={userSigner}
+              provider={mainnetProvider}
+              address={address}
+              blockExplorer="https://etherscan.io/"
+            />
+            */}
+        </Route>
+        <Route path="/subgraph">
+          <Subgraph
+            subgraphUri={props.subgraphUri}
+            tx={tx}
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -6,7 +6,6 @@
```
carletex, line 6: Not sure about the changes in this file. Just a auto-formatting thing?

I'd remove the tweaks to this file.

packages/react-app/src/App.jsx
```diff
@@ -1,13 +1,6 @@
 import { Button, Col, Menu, Row } from "antd";
 import "antd/dist/antd.css";
-import {
```
carletex, line 3: I think you might have a custom prettier/formatting settings in your IDE (different from `packages/react-app/.prettierrc`), because I see some formatting rules not defined in the `.prettierrc` file.

E.g. this line won't go in a single line because the length would be more than > 120 chars (rule: `"printWidth": 120` in the `.prettierrc` file.

packages/react-app/src/App.jsx
```diff
@@ -60,13 +53,14 @@ const DEBUG = true;
 const NETWORKCHECK = true;
 const USE_BURNER_WALLET = true; // toggle burner wallet feature
 const USE_NETWORK_SELECTOR = false;
+const POLL_TIME = 5000;
```
carletex, line 56: `RPC_POLL_TIME` could be a better name.

Also would be nice to explain what this is, and that if you set it to 0 it'll make RPC requests onblock

packages/react-app/src/App.jsx
```diff
@@ -85,15 +80,35 @@ function App(props) {
...
+  }, [provider, mainnetProvider]);
+
+  // load all your providers
+  // const provider = useStaticJsonRPC([
+  //   process.env.REACT_APP_PROVIDER ? process.env.REACT_APP_PROVIDER : targetNetwork.rpcUrl,
+  // ]);
+
+  // console.log("n-provider: ", provider);
+  // const mainnetProvider = useStaticJsonRPC(providers);
+  // console.log("n-mainnetProvider: ", mainnetProvider);
```
carletex, line 107: We don't need this.

packages/react-app/src/App.jsx
```diff
@@ -106,13 +121,24 @@ function App(props) {
     }, 1);
   };
 
-  /* 💵 This hook will get the price of ETH from 🦄 Uniswap: */
-  const price = useExchangeEthPrice(targetNetwork, mainnetProvider);
+  // /* 💵 This hook will get the price of ETH from 🦄 Uniswap: */
+  // console.log("n-isLoaded: ", isLoaded);
+  // console.log("n-isLoaded ? 5000 : 500: ", isLoaded ? 5000 : 500);
```
carletex, line 126: Remove

packages/react-app/src/App.jsx
```diff
@@ -77,6 +71,7 @@ function App(props) {
 
   const [injectedProvider, setInjectedProvider] = useState();
   const [address, setAddress] = useState();
+  const [isLoaded, setIsLoaded] = useState(false);
```
carletex, line 74: What is the use for this?

I see you are doing `isLoaded ? POLL_TIME : 500,` below, but not sure what is the intention with it.

packages/react-app/src/hooks/useStaticJsonRPC.js
```diff
@@ -2,32 +2,75 @@ import { useCallback, useEffect, useState } from "react";
 import { ethers } from "ethers";
```
carletex, line 2: I think we can make the changes on this file a bit more straight-forward.

We can leave as it was before and only adding a `localProvider = null` param to the `useStaticJsonRPC` hook. Then, inside the hook:

- We check if localProvider exists and is mainnet. 
- If is mainnet => init the provider useState with localProvider.
- If is NOT mainnet => init useState with null (as it was)
- On the useEffect, we don't call handleProviders if localProvider exists and is mainnet. 


This way the hook will return localProvider if mainnet, and if not it'll initialize a new provider.

carletex (commented):
Hey Naim, thanks for this!

I think it's a good starting point. I left some review comments so we can keep iterating.

Also it's important to check the difference on RPC requests before & after this changes. (You can check on your brower devtools => network tab)

packages/react-app/src/App.jsx
Naim-Bijapure: ok will update

packages/react-app/src/App.jsx
Naim-Bijapure: will remove it

packages/react-app/src/App.jsx
Naim-Bijapure: yes

packages/react-app/src/App.jsx
Naim-Bijapure: ya, i think it's unnecessary I added it while building the logic. 
the intention was to add 500ms of time at the start to load price or balance etc quickly

packages/react-app/src/hooks/useStaticJsonRPC.js
Naim-Bijapure: sure let me think on this

README.md
Naim-Bijapure: ya ignore this

packages/react-app/src/hooks/useStaticJsonRPC.js
```diff
@@ -25,9 +25,23 @@ export default function useStaticJsonRPC(urlArray) {
...
   useEffect(() => {
-    handleProviders();
+    //  localProvider is null  and urlArray's length is 1 then load a first  localProvider
+    if (localProvider === null && urlArray.length === 1) {
+      handleProviders();
+    }
+
+    //  localProvider is not  null check chain id if it is mainnet id  then set provider
+    if (localProvider !== null && localProvider?._network.chainId === 1) {
+      setProvider(provider);
```
carletex, line 35: You are setting an empty provider here.

It should be `setProvider(localProvider)`.

Also the logic on this useEffect is a bit confusing. Not sure if we need those conditions. I think it could be something like:

```js
    if (localProvider && localProvider?._network?.chainId === 1) {
      setProvider(localProvider);
      return;
    }

    handleProviders();
```

packages/react-app/src/App.jsx
```diff
@@ -157,17 +159,21 @@ function App(props) {
...
-  });
+  // useOnBlock(mainnetProvider, () => {
+  //   console.log(`⛓ A new mainnet block is here: ${mainnetProvider._lastBlockNumber}`);
+  // });
 
   // Then read your DAI balance like:
-  const myMainnetDAIBalance = useContractReader(mainnetContracts, "DAI", "balanceOf", [
-    "0x34aA3F359A9D614239015126635CE7732c18fDF3",
-  ]);
+  const myMainnetDAIBalance = useContractReader(
```
carletex, line 167: We also need to add the RPC_POLLING_TIME to the `useGasPrice` and `useExchangeEthPrice` hooks in this file.

carletex (commented):
Hey, left some review comments.

Some other things:
- Check the `useBalance` in other files (you also need to add the polling time there)
- The providers have a pollingInterval prop that you can also set like `localProvider.pollingInterval = 50000;`. This avoids the continuous  call to `eth_getBlockNumber`. Not sure the side effects of this tho.

packages/react-app/src/hooks/useStaticJsonRPC.js
Naim-Bijapure: ohh sry its a type :)

packages/react-app/src/App.jsx
```diff
@@ -62,11 +62,12 @@ const USE_BURNER_WALLET = true; // toggle burner wallet feature
 const USE_NETWORK_SELECTOR = false;
 
 const web3Modal = Web3ModalSetup();
+const RPC_POLL_TIME = 5000;
 
 // 🛰 providers
 const providers = [
   "https://eth-mainnet.gateway.pokt.network/v1/lb/611156b4a585a20035148406",
-  `https://eth-mainnet.alchemyapi.io/v2/${ALCHEMY_KEY}`,
+  `https://eth-mainnet.alchemyapi.io/v2/${ ALCHEMY_KEY }`,
```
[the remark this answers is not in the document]
Naim-Bijapure: not sure have checked that its following the rules . look like eslint plugin also overriding it

packages/react-app/src/App.jsx
```diff
@@ -65,9 +67,9 @@ const web3Modal = Web3ModalSetup();
 
 // 🛰 providers
 const providers = [
-  "https://eth-mainnet.gateway.pokt.network/v1/lb/611156b4a585a20035148406",
+  // "https://eth-mainnet.gateway.pokt.network/v1/lb/611156b4a585a20035148406",
```
carletex: We don't want to comment this out

packages/react-app/src/App.jsx
```diff
@@ -107,7 +110,7 @@ function App(props) {
   };
 
   /* 💵 This hook will get the price of ETH from 🦄 Uniswap: */
-  const price = useExchangeEthPrice(targetNetwork, mainnetProvider);
+  const price = useExchangeEthPrice(targetNetwork, mainnetProvider, RPC_POLL_TIME);
 
   /* 🔥 This hook will get the price of Gas from ⛽️ EtherGasStation */
   const gasPrice = useGasPrice(targetNetwork, "fast");
```
carletex, line 116: We need to set the poll time also here.

packages/react-app/src/constants.js
```diff
@@ -9,6 +9,8 @@ export const BLOCKNATIVE_DAPPID = "0b58206a-f3c0-4701-a62f-73c7243e8c77";
 
 export const ALCHEMY_KEY = "oKxs-03sij-U_N0iOlrSsZFr29-IqbuF";
 
+export const RPC_POLL_TIME = 3000;
```
carletex: Maybe 5000 as default?

Also as add a comment above telling what that constant is doing.

have you tried with 0? That should do RPC request on block, as it was before.

packages/react-app/src/hooks/useStaticJsonRPC.js
```diff
@@ -9,14 +9,17 @@ const createProvider = async url => {
...
   const [provider, setProvider] = useState(null);
 
   const handleProviders = useCallback(async () => {
     try {
       const p = await Promise.race(urlArray.map(createProvider));
       const _p = await p;
 
+      // set localProviders internal polling interval
+      // _p.pollingInterval = 50000;
+
```
carletex, line 22: Why is this commented? 

The default (inside ethers providers) is 4000. So if RPC_POLLING_TIME is bigger, we cant set it, if not, leave the default

PR #894, opened 2022-09-25
Fix GitPod Issue

codenamejason:
Fix for GitPod error/issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,11 +68,11 @@
       "**/hardhat-ts/**"
     ]
   },
-  "dependencies": {},
-  "devDependencies": {
+  "dependencies": {
     "husky": "^7.0.2",
     "lint-staged": "^13.0.3"
   },
+  "devDependencies": {},
   "lint-staged": {
     "packages/react-app/src/**/*.{js,jsx}": [
       "prettier --write",


codenamejason:
this fixes the gitpod issue per @technophile-04 recommendation 🙌

codenamejason:
Seems like `Gitpod` was not installing `husky` because it was placed in `dev dependencies`. Adding `husky` to `main dependencies` instead of `dev dependencies` solved this issue. 

Here is the Gitpod link to my repo : 

[Open in Gitpod](https://gitpod.io/#github.com/technophile-04/scaffold-eth)

Alternate solution: Maybe we can ran `yarn add -D husky` after `yarn install` in `.gitpod.yaml`. No idea about it, will explore this.

PS: Don't know why doing `yarn install` even before adding `husky` to `main dependencies` updated `yarn.lock`.The merege conflicts in`yarn.lock` are due to that reasons I think. 

Open to any suggestions/improvements 🙌

damianmarti:
> From @technophile-04 =>
> 
> Seems like `Gitpod` was not installing `husky` because it was placed in `dev dependencies`. Adding `husky` to `main dependencies` instead of `dev dependencies` solved this issue.
> 
> Here is the Gitpod link to my repo :
> 
> [Open in Gitpod](https://gitpod.io/#github.com/technophile-04/scaffold-eth)
> 
> Alternate solution: Maybe we can ran `yarn add -D husky` after `yarn install` in `.gitpod.yaml`. No idea about it, will explore this.
> 
> PS: Don't know why doing `yarn install` even before adding `husky` to `main dependencies` updated `yarn.lock`.The merege conflicts in`yarn.lock` are due to that reasons I think.
> 
> Open to any suggestions/improvements raised_hands

What was the problem for @technophile-04 in Gitpod?

technophile-04:
Hii @damianmarti  sorry for the late reply, So currently you can't open the scaffold-eth main repo with Gitpod. It gives this error : 
Screenshot 2022-09-27 at 5 23 23 PM

To fix this we need to move `husky` from `dev dependencies` to `main dependencies` since Gitpod is not able to recognize it. I and @codenamejason created the PR at the same time, since mine had some merge conflicts in `yarn.lock` I closed my PR. we can merge this PR  🙌.

here is an example https://gitpod.io/#github.com/technophile-04/scaffold-eth which uses the proposed solution in this PR and works.

damianmarti:
thanks @technophile-04 for the data!

is there a way to avoid running the husky post checkout hook when deploying into gitpod?

I personally don't like this husky post checkout hook running yarn install. Sometimes I just want to switch branches to see at some code and I don't want to install the libraries.

I vote to leave the package.json as is was and removing the post checkout hook hook ;-)

cc/ @codenamejason

codenamejason:
Same opinion here @damianmarti the user could run to learn the process instead of automatic.

carletex:
> I personally don't like this husky post checkout hook running yarn install. Sometimes I just want to switch branches to see at some code and I don't want to install the libraries.

+1

pre-commit linting hook is handy tho.

PR #896, opened 2022-09-26
Sensible defaults for RPC calls

carletex:
This PR aims to reduce the RPC calls made by scaffold-eth on a clean install (around 60/80 per minute) by providing some sensible defaults on polling times (vs onBlock) and reusing providers (if localProvider is mainnet).

There are two new constants (with sensible defaults) that config this behavior: `LOCAL_RPC_POLL_TIME` and `MAINNET_RPC_POLL_TIME`. 

On the localhost (hardhat chain) the polling is disabled (uses onBlock) to keep allowing a fast feedback loop. (We don't care about RPC requests to the localnode)

Great work @Naim-Bijapure & @technophile-04


carletex:
Not sure about the failing GitHub Action test on the "actions/first-interation" task

packages/react-app/src/App.jsx
```diff
@@ -107,10 +113,10 @@ function App(props) {
   };
 
   /* 💵 This hook will get the price of ETH from 🦄 Uniswap: */
-  const price = useExchangeEthPrice(targetNetwork, mainnetProvider);
+  const price = useExchangeEthPrice(targetNetwork, mainnetProvider, 1000);
```
damianmarti, line 116: Is ok 1000 here? Maybe MAINNET_RPC_POLL_TIME?

carletex:
Still iterating and testing. Some findings:

## Finding 1
This is from eth-hooks v2. Hook [useOnRepetition](https://github.com/scaffold-eth/eth-ui/blob/releases/v2/src/useOnRepetition.ts). This hook is being used on all the hooks that uses polltime (useBalance, useContractReader, etc).

image

**Conclusion**: `LOCAL_RPC_POLL_TIME = 5000` doesn't make sense.
**Action**: Use a bigger time? 30000 maybe?

## Finding 2
We already know that when creating a new ethers provider (new ethers.providers.StaticJsonRpcProvider()), it will result in a RPC call `eth_blockNumber` number every 4000 ms.

image
*From ethersjs base-provider*

We can change that base provider polling time. On this PR we are hardcoding it:

image

If you set pollingInterval to 60000 (1 minute) it will make 15 times less RPCs calls. **The thing is:** if a hook is executed without a polling time (e.g. useBalance) it will make a RPC call (e.g. eth_getBalance) for **EVERY** block (even if the provider checks for new block every 60 secs). This is because when polling time is not defined for a hook, it will run onBlock.

Detailed example: (pollingInterval = 60000, useBalance() with no polling time):

**Conclusion**: Increasing the internal provider pollingInterval won't make a difference in hooks without polling time
**Actions**: Don't hardcode the pollingInterval. Use polling time on hooks (we are already doing that)

## Finding 3

When using the internal provider pollingInterval + hooks with polling time, it works as expected. Except one thing.

If you set 30000 ms for a polling time for a hook... it will wait 30 seconds to get the data. It should get the data instantly and then wait 30 s to poll it again.

This happens only for some hook (like useExchangeEthPrice). Others (useBalance) works fine.

This is something we need to fix on eth hooks v2. There is an option to this initial call (leadingTrigger prop) that it's not present in useExchangeEthPrice. Opened a PR there to see if @ShravanSunder can add it (or there is a reason for not having that prop)

image

**Action**. If the eth-hooks PR gets merged, updated the eth-hooks version.

technophile-04:
Thanks so much @carletex !! this is really very helpful 🙌, Here's my understanding and gains : 
**Finding 1:** lol `LOCAL_RPC_POLL_TIME = 5000` don't make any sense because since `5000 < 10000` it will default to `10000` and makes `LOCAL_RPC_POLL_TIME = 5000` completely useless, will try to test it with 30000 .

**Finding 2:** Is the detailed example complete 😅 ? According to my understanding of above description will try to expand it : 
Detailed example: (pollingInterval = 60000, useBalance() with no polling time): Lets say in 60 sec there are `4 block mined `
so instead of making **1** `eth_getBalance`  after 60 seconds it will make **4** `eth_getBalance` calls instead because it called  for **Every** block. 
So by increasing the `pollingInterval` we are just saving some `eth_blocknumber` calls, please let me know if I am wrong somewhere. What if we decrease  it 15sec or 12sec(avg block mined time) need to test it what difference this makes ?

**Finding 3 :** I really was having this doubt, why the `ETH price` was not updating instantly and this seems to solve my doubt completely 🙌.  
Since this is not used at many places on fresh `yarn install` of scaffold-eth can we wrap this hook around **our** hook which make first calls instantly and then does it according to pollTime...for time being? I have no idea how to achieve it just thinking out loud.

packages/react-app/src/App.jsx
carletex: Fixed, thanks! ;)

carletex:
Thanks for the review & testing @technophile-04 

**Finding 2**: lol. Yeah I forgot to write the example, sorry haha But it's exactly how you describe it. 

Pushed a new commit:
- Unify RPC constants in one
- Increase default poll time to 30000 (Finding 1)
- De-hardcode the pollingInterval time

Finding 3: I'm going try to get that fixed on eth-hooks v2. Will create a PR later.

Can you test all this please? Thanks Shiv!

carletex:
@technophile-04 Another finding!

I set up a minimal version of a **freshly cloned** scaffold-eth: basically removing most of the stuff from App.jsx so I could test things in isolation.

```
  const mainnetProvider = useStaticJsonRPC(providers);
  const mainnetBalance = useBalance(mainnetProvider, "0x34aA3F359A9D614239015126635CE7732c18fDF3", 15000);
```

What I expected from this:
- eth_blockNumber calls every 4 seconds (default pollingInterval in ethers providers) to get the blockNumber
- eth_getBalance calls every 15 seconds (polling time that we specified)

What I got:
- eth_getBalance calls every 15 seconds
- NO `eth_blockNumber` calls!

This makes me think that we have something subscribed to the provider (not using poll time) that make the provider call eth_blockNumber. We shouldn't need those calls if everything has a pollTime.

I'll try to find it!

technophile-04:
Interesting 🤔 !! Even I tried it out and there are no `eth_blockNumber` calls! I'll also try my best to see if I find something...just started testing it out

carletex:
@technophile-04 Just realized that with the latest changes, we are not calling `eth_blockNumber` anymore. So I think we don't need to set the internal pollingTime. I removed it.

Please test :)

The only think we need before merging is fixing the polling time for `useExchangeEthPrice`. I'll open a PR on the eth-hooks repo. I'll keep you posted!

carletex:
Create the PR on eth-hooks (eth-ui monorepo) https://github.com/scaffold-eth/eth-ui/pull/243

codenamejason:
> Not sure about the failing GitHub Action test on the "actions/first-interation" task
> 
> cc @codenamejason

I disabled it for now

technophile-04:
@carletex really liked the new changes they makes it a lot cleaner, simple, and easy to wrap your head around, this is so awesome !! 

### Test on localchain : 

Everything seems to work as expected and UX is also the same as `master` branch since pollTime is set to `0`.  Here are some basic test which I could think of based on the changes.
- [x] After grabbing funds from faucet, Balance should updates quickly 
- [x] The `eth_getBlance` call to  localchain should be made instantly and **only** when new block is added. 
- [x] The `eth_getBlance` call to  mainnet should be made instantly and then after every 30sec. 
- [x] The `eth_call` for `daiBlance` should be made instantly and then after every 30sec. 
- [ ] `eth_call` for `ethPrice` from should be made instantly and then after every 30sec. (will be fixed soon 🙌)
- [x] setting the purpose from `Example UI` should update the purpose quickly after 4 sec. 
- [x] `eth_call` to read `purpose` should be made **only** when new block is added. 
- [x] Check if the Events reading is done quickly 

Sorry If I missed some, please do let me know, still testing and will be updating here...

### Test on goerli :
- [x] Nothing breaks if you try to set `RPC_POLL_TIME = 0`  .
- [x] `eth_getBlance` calls to goerli node should be made instantly and then after every 30sec.
- [x] All the mainnet calls should be made instantly and after ever 30sec . 
- [x] `eth_call` for `purpose` is made after every 30sec (using RPC_POLL_TIME=30000).
- [x] If `RPC_POLL_TIME=0` `eth_call` are only made when there is addition of newBlock.
- [x] Events also remain in sync and updates quickly (🙌 since pollingInterval for rpc was removed)

Still testing and will keep it updated...

### Some findings/can be improved I think : 
If we forgot to update the `RPC_POLL_TIME` when on a different chain, and when you connect your metamask wallet address,  it updates the balance in  navbar after a `30sec`. Since when you connect the page is not refreshed thats why `eth_balance` is not made instantly and is made after `30sec` which I think is a bad UX, its a rare case though just putting it here. 
[nonchalant-joke.surge.sh](http://nonchalant-joke.surge.sh/) here you can  see the delay of updation of balance after connecting to the wallet.

Just a suggestion we can completely ignore it since will be updating it in readme and at some good place. We can just remind with a info like this which run only once when you are using chain other than localhost and the dev environment is not production : 
infoSS|

### Sharing some comparisons of no.of rpc calls **per minute**  with `master` branch on fresh `yarn install` 🤩 : 
master branch (all calls)                    |  rpc-calls-sensible-defaults (all calls)
:-------------------------:|:-------------------------------------------------------:
allCalls | sensibleAllCal

master branch (mainnet calls)                    |  rpc-calls-sensible-defaults (mainnet calls)
:-------------------------:|:-------------------------------------------------------:
mainnetRpc calls | sensibleMainnetRPC

packages/react-app/src/constants.js
```diff
@@ -9,6 +9,17 @@ export const BLOCKNATIVE_DAPPID = "0b58206a-f3c0-4701-a62f-73c7243e8c77";
 
 export const ALCHEMY_KEY = "oKxs-03sij-U_N0iOlrSsZFr29-IqbuF";
 
+/*
+Decrease the number of RPC calls. This value is being passed to the provider
+internal polling time (it'll look for new blocks after this time) and hooks
```
technophile-04, line 14: also, we can update this line 🙌

technophile-04, line 14: also, we can update this 🙌

carletex: Good catch! Fixed on 14822892abc4dad9b417af569715741e9979192d

carletex:
Hey @technophile-04 

Amazing!! Thank you for the detailed and thorough response :)

> eth_call for ethPrice from should be made instantly and then after every 30sec. (will be fixed soon raised_hands)

PR with the fix was merged on eth-hooks & uploaded to npm. I updated the version here and now the eth price loads after a few seconds :)

> here you can see the delay of updation of balance after connecting to the wallet.

Good catch on this. Yes, I think it's bad UX having to wait 30 seconds after you connect your wallet to see your Balance.

The thing is: the problem is not the 30 sec RPC_POLL_TIME. The problem is on the `useBalance` hook from eth-hooks. **It should refresh when the address change**. useBalance calls useOnRepetition under the hood, and there is where the problem is.

To test, I copied over the JS from useBalance & useOnRepetition functions (from node_modules) to the project, import those ones (instead of using eth-hooks) and made some tweaks. At the end of useOnRepetition you have:

```js
useEffect(() => {
    if (options.leadingTrigger && callFunctionWithArgs != null && leadingCall?.current === true) {
      leadingCall.current = false;
      callFunctionWithArgs();
    }
  }, [options.leadingTrigger, callFunctionWithArgs]);
```

We need to trigger a call to that useEffect when the address changed. To do that, what I did:

**1. Pass address as property on `useBalance`**

From:
```js
  useOnRepetition(
    pollBalance,
    { pollTime, provider, leadingTrigger: address != null && address !== "" && provider != null },
    provider,
    address,
  );
```

To:
```js
  useOnRepetition(
    pollBalance,
    { pollTime, provider, leadingTrigger: address != null && address !== "" && provider != null, address },
    provider,
    address,
  );
```

**2. Trigger useEffect On useOnRepetition:**

From
```js
  useEffect(() => {
    if (options.leadingTrigger && callFunctionWithArgs != null && leadingCall?.current === true) {
      leadingCall.current = false;
      callFunctionWithArgs();
    }
  }, [options.leadingTrigger, callFunctionWithArgs]);

```

```js
  useEffect(() => {
     ////////// Remove leadingCall ref => NOT SURE ABOUT SIDE EFFECTS
    if (options.leadingTrigger && callFunctionWithArgs != null) {
      callFunctionWithArgs();
    }
   ////////// Add options (vs options.leadingTrigger) to run this on any options change
  }, [options, callFunctionWithArgs]);
```

[Here](https://github.com/scaffold-eth/eth-ui/tree/releases/v2/src) you can see the hooks for eth-hooks v2 (also in your local noode_modules).

Working version: https://stupendous-airport.surge.sh/

This is how it should work. But not sure about the side effect of the changes.

Would love to have your feedback and testing on this.

technophile-04:
@carletex thanks so much for all the references and suggestions 🙌, I tried testing the approach and saw that number of localProvider rpc calls(using goerli as network) increased a lot. 
rpc-withOptions


**Possible reason** : 
```diff
// inside useBalance hook
 useOnRepetition(
    pollBalance,
+    { pollTime, provider, leadingTrigger: address != null && address !== "" && provider != null, address },
    provider,
    address,
  );
```
```diff
// inside useOnRepetition hook
  useEffect(() => {
     ////////// Remove leadingCall ref => NOT SURE ABOUT SIDE EFFECTS
    if (options.leadingTrigger && callFunctionWithArgs != null) {
      callFunctionWithArgs();
    }
   ////////// Add options (vs options.leadingTrigger) to run this on any options change
+  }, [options, callFunctionWithArgs]);
```

since we are using `options` _**object**_ as our dependency, and since react uses [referential equality](https://dev.to/vicnovais/understanding-referential-equality-in-reacts-useeffect-2m7o) for dependencies in useEffect, the `options`object changes every time and it causes the useOnRepetition's **useEffect** to fire up. 

**Possible Solution (adding new useEffect)** : 
```js
// useOnRepetition hook
export const useOnRepetition = (callback, options, ...args) => {
    ...
    const argsAddress = args !== null && args !== void 0 && args.length >= 2 ? args[1] : "";
    useEffect(() => {
        if (callFunctionWithArgs != null) {
          callFunctionWithArgs();
        }
      }, [argsAddress]);
};
```
Using `string` in dependency array so that it only triggers `useEffect` only when **address** _**value**_ is changed. 
Here are the localProvider rpc calls with above changes : 
 rpc-with-onlyAddress
 
 I tried changing different useCallback, useEffects dependencies with `address` but this seems more feasible with less RPC call,  please do let me know if I missed something or something I am wrong about  🙌
 
Here is the link to [repo](https://github.com/technophile-04/scaffold-eth/tree/rpc-test) for the test. custom useBalance and useRepetition hook added [here](https://github.com/technophile-04/scaffold-eth/blob/d687f190ba7313834f76a845baecdd1694c5fb43/packages/react-app/src/hooks/useBalance.js#L104). It also has previouse apporoch commented above for the test. I have commented out `Debug Contract` and `Events` component in `Example UI` just to focus on necessary rpc request .

carletex:
Hey @technophile-04 thanks for this ;)

As we talked before through DMs there is no easy way to do this without making a bunch of changes on eth-hooks. I've been playing around a bit, and came with a hacky solution with minimal changes in eth-hooks.



Changes on useOnRepetition: I'm playing with the fact that the useEffect we want to trigger, has options.leadingTrigger as deps, and useBalance set it to null if !address and to true if it's defined.

**Before**
```js
    useEffect(() => {
        if (options.leadingTrigger && callFunctionWithArgs != null && (leadingCall === null || leadingCall === void 0 ? void 0 : leadingCall.current) === true) {
            leadingCall.current = false;
            callFunctionWithArgs();
        }
    }, [options.leadingTrigger, callFunctionWithArgs]);
```

**After**
```js
    useEffect(() => {
        if (options.leadingTrigger && callFunctionWithArgs != null) {
            callFunctionWithArgs();
        }
    }, [options.leadingTrigger, callFunctionWithArgs]);
```

It's removing the leadingCall ref. Again, not sure about side effects.

Then in our Balance component (bit of a hacky solution):

```js
  // New state for address
  const [address, setAddress] = useState(props.address);

  useEffect(() => {
    if (!address) {
      setAddress(props.address);
      return;
    }

    if (props.address !== address) {
      // This will set the address to null => useBalance hook will have leadingTrigger to false => will make useOnRepetition useEffect to re-run.
     // This will also re-run this useEffect :)
      setAddress(null);
    }
  }, [props.address, address]);

  const balance = useBalance(props.provider, address, localProviderPollingTime);
```

I see 4 possible solutions (in order of preference, more or less)
1. Use the tweaks I suggested in [here](https://github.com/scaffold-eth/scaffold-eth/pull/896#issuecomment-1265166247) (fixing the option object issue you mentioned). Minimal changes in eth-hooks, no changes in SE.
2. Re-write eth-hooks useBalance hook (e.g. return a re-fetch function). This will also imply changing all the useBalance uses in SE.
3. Write our own useBalance hook inside the SE repo. (without using useOnRepetition tho)
4. Use a hacky solution with minimal changes in eth-hooks + change in SE

What do you think? 

Also, happy to hear more solutions that you might come up with!

carletex:
Just noticed that if we remove the leadingCall ref logging... `callFunctionWithArgs` is being called on every polltime. So we can't remove it.

It can be fixed with this:

```js
  useEffect(() => {
    if (!options.leadingTrigger) {
      // Reset leading call
      leadingCall.current = true;
    }

    if (
      options.leadingTrigger &&
      callFunctionWithArgs != null &&
      (leadingCall === null || leadingCall === void 0 ? void 0 : leadingCall.current) === true
    ) {
      leadingCall.current = false;
      callFunctionWithArgs();
    }
  }, [options.leadingTrigger, callFunctionWithArgs]);
```

carletex:
Ok @technophile-04 

Just updated the eth-hooks version and everything seems to work. This should be ready to go :)

Could you double check?

Good work Shiv!

technophile-04:
Tested it out both locally and goerli and everything works as we intended it to be 🙌, also rested all the things mentioned in this [comment](https://github.com/scaffold-eth/scaffold-eth/pull/896#issuecomment-1264483747), 
- [X] `eth_call` for ethPrice from should be made instantly and then after every 30sec.
- [X] The balance should be instantly re-fetched when you change the address.

Both the above two things are solved and works great now, 

Regarding the `useRepetition` changes as mentioned in this [PR](https://github.com/scaffold-eth/eth-ui/pull/244) there are two types of hooks that were using `useOnRepetition` and new changes don't disturb any of those, even `useContractReader` is not affected since it passes `array` as args which is ignored by new changes, checkout out this [PR](https://github.com/scaffold-eth/eth-ui/pull/244)  `Test` section for more context. 

Everything seems fine to me 🙌

Test Repo : https://github.com/technophile-04/scaffold-eth/tree/rpc-test-main
Demo : https://handsomely-noise.surge.sh/exampleui

PS : The purpose in the demo will be updated after `30sec`as I didn't change  `RPC_POLL_TIME ` for the test,  but the balance updates as soon as you connect the wallet.

technophile-04 (approved):
🚀🚀

codenamejason (commented):
Testing this on my local machine and noticed a couple things:

To reproduce:
1. run `yarn chain`
2. run `yarn start`
3. navigate to the Debug Contracts tab
4. It shows loading... as expected but why is there an address yet? This is causing some side effects to happen and cause an error (see screens).

Expected: The user should be able to navigate to this tab and have proper status and no errors whether contracts have been deployed or not.

Note: Not sure if this is new behavior (will have to check `master` branch behavior) and may have nothing to do with this PR 🫣

Screen Shot 2022-10-13 at 12 26 19 PM

Screen Shot 2022-10-13 at 12 26 32 PM

Screen Shot 2022-10-13 at 12 26 07 PM

codenamejason (approved):
LGTM

technophile-04:
Hii @codenamejason, I tried reproducing this with fresh scaffold-eth install and it seem to work : 
Screenshot 2022-10-14 at 2 37 50 PM

This behavior happens when : 
1. run `yarn chain`
2. run `yarn deploy` (deploys the contract and creates `hardhat_contracts.json` in `react-app/src/contracts`, adds deployed `contract address`,`abi`  in `hardhat_contracts.json`)
3. kill the local chain 
4. run `yarn chain` (This is the new chain and the previous deployed contract is not present on this chain)
5. run `yarn start` (It first creates `hardhat_contracts.json` if its not present, with an empty object,  in our case its already present and then spins up the react app)

Now since `hardhat_contracts.json` has `contract address` and everything init (but all this is not present in new local chain) that's why `Contract` component is displaying the address of the previous deployed contract but not loading the functions and all because its not present on actual current chain. hope it makes sense 😅

Possible Solution : 
There is this hooked used in `Contract` component `useContractExistsAtAddress` which checks if contract exists and displays the loading which you shared  with some additions informations if its not present, maybe we can use it

damianmarti:
LGTM too!

Awesome improvements!!!

PR #897, opened 2022-09-27
updated the logic to use no polling when on localhost and increased pollTime for mainnet calls

technophile-04:
## <u> **Current situation #896 :** </u>
- When you send funds through the faucet it takes around `5 sec` to reflect since `RPC_POLL_TIME = 5000` which I think is a bit slow for local development.

- If you try setting `purpose` through the `ExampleUI` tab it again takes `5 sec` to update the purpose which is slow. 

- Another problem is let's suppose a builder creates a new `Contract` and tries to read from the contract  using `useContractReader` hook and fails to pass `pollInterval` in arguments, Now if the value updates it will take **`50 sec`** to reflect that in `UI` due to ` _p.pollingInterval = 50000;` in `useStaticJsonRPC.js`. Since `useContractReader` updates the values when new block is added when `pollInterval` argument is not passed or it is passed as `0` and since `_p.pollingInterval` is set to `50000` it will check for new block after `50000` .

-  In `App.jsx`,  `yourMainnetBalance` and `myMainnetDAIBalance` are using `Mainnet RPC`,  having `pollTime` of `5 sec` 
and this is only used to show the usage of `hooks` and are only `console.logged` in `useEffect`. It creates a load on `Mainnet RPC` used .

## <u> **Proposed Solution :** </u>
- Update the logic in `useStaticJsonRPC.js` to not use ` _p.pollingInterval = 50000;` when using `localhost` also updated `RPC_POLL_TIME` variable wherever used to be `0` when you are on `localhost`. This helps the ui reflect quickly .

- Increased the `PollTime` for `yourMainnetBalance` and `myMainnetDAIBalance` which decreases the request to `Mainneet RPC's` by around `7-8` **`calls per min`**
  PS : Created a constant `MAINNET_RPC_POOL_TIME`, Instead we can directly put in value as parameter.

### Before : 

All Fetch/XHR Calls            |  Showing only calls to Mainnet(Filtered using domain)
:-------------------------:|:-------------------------------------------------------:
Screenshot 2022-09-27 at 4 49 20 PM |  Screenshot 2022-09-27 at 4 44 49 PM

## After Changes : 

All Fetch/XHR Calls            |  Showing only calls to Mainnet(Filtered using domain)
:-------------------------:|:-------------------------------------------------------:
Screenshot 2022-09-27 at 4 54 14 PM 1 |  Screenshot 2022-09-27 at 4 56 53 PM

PS : I think we should decrease the ` _p.pollingInterval = 50000;` to `30000` i.e around 2 blocks or maybe we can mention somewhere to decrease `_p.pollingInterval = 50000;` because if someone uses `useContractReader` without passing  Pollinterval argument and using chain other than `localchain`, he/she will find it hard to debug like I barely anytime time pass pollInterval to `useContractReader` but I may be wrong 😅. 

Sorry for my bad explanations 😅 , would try my best to explain again if there are any doubts 🙌

### DIFF
--- a/packages/react-app/src/components/Balance.jsx
+++ b/packages/react-app/src/components/Balance.jsx
@@ -1,6 +1,6 @@
 import React, { useState } from "react";
 import { useBalance } from "eth-hooks";
-import { LOCAL_RPC_POLL_TIME } from "../constants";
+import { getRPCPollTime } from "../helpers";
 
 const { utils } = require("ethers");
 
@@ -33,11 +33,7 @@ const { utils } = require("ethers");
 export default function Balance(props) {
   const [dollarMode, setDollarMode] = useState(true);
 
-  let RPC_POLL_TIME = LOCAL_RPC_POLL_TIME;
-
-  if (props.provider && props?.provider._network.chainId === 31337) {
-    RPC_POLL_TIME = 0;
-  }
+  let RPC_POLL_TIME = getRPCPollTime(props.provider);
 
   const balance = useBalance(props.provider, props.address, RPC_POLL_TIME);
   let floatBalance = parseFloat("0.00");
--- a/packages/react-app/src/components/Balance.jsx
+++ b/packages/react-app/src/components/Balance.jsx
@@ -1,6 +1,6 @@
 import React, { useState } from "react";
 import { useBalance } from "eth-hooks";
-import { RPC_POLL_TIME } from "../constants";
+import { LOCAL_RPC_POLL_TIME } from "../constants";
 
 const { utils } = require("ethers");
 
@@ -32,6 +32,12 @@ const { utils } = require("ethers");
 
 export default function Balance(props) {
   const [dollarMode, setDollarMode] = useState(true);
+
+  let RPC_POLL_TIME = LOCAL_RPC_POLL_TIME;
+
+  if (props.provider && props?.provider._network.chainId === 31337) {
+    RPC_POLL_TIME = 0;
+  }
 
   const balance = useBalance(props.provider, props.address, RPC_POLL_TIME);
   let floatBalance = parseFloat("0.00");
--- a/packages/react-app/src/constants.js
+++ b/packages/react-app/src/constants.js
@@ -9,7 +9,11 @@ export const BLOCKNATIVE_DAPPID = "0b58206a-f3c0-4701-a62f-73c7243e8c77";
 
 export const ALCHEMY_KEY = "oKxs-03sij-U_N0iOlrSsZFr29-IqbuF";
 
-// LOCAL_RPC_POLL_TIME as 5 seconds to slower the rpc calls at start
+/* To decrease the number of RPC calls `LOCAL_RPC_POLL_TIME` is set to `5000ms` which can be passed to hooks which has pollTime like (useContractReader, useBalance etc) .
+
+NOTE : If you are not using local-hardhat chain and pass poolTime as 0 to hooks which uses pollTime they defaults to use `onBlock`. Checkout `src/hooks/useStaticJsonRPC.js` where provider pollingInterval is set to 30000ms i.e it will be checking for blocks after 30secs
+
+*/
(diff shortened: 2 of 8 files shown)


technophile-04:
To test pt.3 in Current Situation(which is #896) , we can just remove `pollInterval`  argument in `App.jsx` at `line 177` where we are using `useContractReader` to read `purpose`

packages/react-app/src/hooks/useStaticJsonRPC.js
```diff
@@ -17,8 +18,10 @@ export default function useStaticJsonRPC(urlArray, localProvider = null) {
       const p = await Promise.race(urlArray.map(createProvider));
       const _p = await p;
 
-      // set localProviders internal polling interval
-      _p.pollingInterval = 50000;
+      if (urlArray[0] !== NETWORKS.localhost.rpcUrl) {
+        // set localProviders internal polling interval, if its not localhost
+        _p.pollingInterval = 50000;
```
carletex, line 23: Yep, agree that `30000` is better here.

packages/react-app/src/constants.js
```diff
@@ -9,8 +9,10 @@ export const BLOCKNATIVE_DAPPID = "0b58206a-f3c0-4701-a62f-73c7243e8c77";
 
 export const ALCHEMY_KEY = "oKxs-03sij-U_N0iOlrSsZFr29-IqbuF";
 
-// RPC_POLL_TIME as 5 seconds to slower the rpc calls at start
-export const RPC_POLL_TIME = 5000;
+// LOCAL_RPC_POLL_TIME as 5 seconds to slower the rpc calls at start
+export const LOCAL_RPC_POLL_TIME = 5000;
+
+export const MAINNET_RPC_POLL_TIME = 30000;
```
carletex, line 15: Can we add / change comments here? Let's explain a little bit better what this does.

We could also mention something about the pollingInterval thing in `useStaticJsonRPC.js`.

carletex, line 15: Also: Does everything works correctly if we set it to 0? My expectation would be that if 0 => every thing work "onBlock". Is that the case?

packages/react-app/src/components/Balance.jsx
```diff
@@ -33,6 +33,12 @@ const { utils } = require("ethers");
 export default function Balance(props) {
   const [dollarMode, setDollarMode] = useState(true);
 
+  let RPC_POLL_TIME = LOCAL_RPC_POLL_TIME;
+
+  if (props.provider && props?.provider._network.chainId === 31337) {
+    RPC_POLL_TIME = 0;
+  }
+
```
carletex, line 41: Since we repeat a similar logic in `App.jsx` and `FauceHint.jsx`.. how do you feel about creating a util for it?

E.g. Create a `rpc.js` file inside the `helpers` folder that exports a `getRpcPollTime(provider)` that handle the logic.

carletex (commented):
This is amazing @technophile-04 

Love how you crafted the explanation in the PR. Super helpful! 

Tested it and it works great. Just left some review comments for minor improvements.

packages/react-app/src/constants.js
technophile-04: Tried setting both `LOCAL_RPC_POLL_TIME ` and `MAINNET_RPC_POLL_TIME ` to 0 and it seems to work correctly, Yup if pollTime is passed as zero it defaults to `onBlock`. It's mentioned in eth-hooks docs also to be sure tried it out locally.

Currently if you are on `localhost` the poll time is passed as zero, So I tried mining an empty block to check it makes `eth_call` for `purpose` variable. Here is the test [repo](https://github.com/technophile-04/scaffold-eth/tree/rpc-calls-sensible-defaults-test) for it
Screenshot 2022-09-28 at 4 42 35 PM

packages/react-app/src/components/Balance.jsx
technophile-04: Yes completely agreed,  will for sure keep this in mind next time 🙌

PR #898, opened 2022-09-29
Fix gitpod without adding husky to main dependencies also remove yarn install post-checkout

technophile-04:
### Fix Gitpod, husky post-checkout not found
Updated the `.gitpod.yml` to install `husky`. This seems to solve the problem. 
here is the link to the current gitpod working branch : https://gitpod.io/#github.com/technophile-04/scaffold-eth/tree/fix-gitpod-husky.

### Remove 'yarn install' on post-checkout discussed in #894 
Added a warning to run `yarn install`  instead of running `yarn install`  on the switching branch.

PS : The warning message can be improved 🙌

Screenshot 2022-09-29 at 9 42 34 PM


codenamejason:
@technophile-04 can you check to see why the greetings action is failing now 🙏🏼 here is the PR that merged it #865 let me know if we should just remove it. Was trying to give feedback to new contributors.

technophile-04:
@codenamejason seems like `actions/first-interaction` is having some issue, The issue is mentioned [actions/first-interaction#101](https://github.com/actions/first-interaction/issues/101) here . 

Found the workaround here  [first-interaction/issues/101#issuecomment](https://github.com/actions/first-interaction/issues/101#issuecomment-1262487501), added the commit for the same. Their team has acknowledged and I think they have started working here is the [comment](https://github.com/actions/first-interaction/issues/101#issuecomment-1266053621). 

Till then we can use the above workaround, also should we keep this workaround even after its fixed? So that if something again happens to `actions/first-interaction` it does not fails checks unnecessarily for new PR's ?  

PS : do let me know if I should create a new PR for this issue only .

codenamejason (approved):

PR #911, opened 2022-10-13
fixing 'hexlify' is not defined error

ericcorriel:
Uncommenting the line as-is produces the following error:

Screen Shot 2022-10-13 at 04 19 07


codenamejason (changes requested):
@ericcorriel can you remove the changes to the two css files they are in #914

ericcorriel:
@codenamejason whoops–sorry about the css files; missed the initial request

codenamejason:
It's OK we're going to archive this in favor of scaffold eth 2

PR #912, opened 2022-10-13
fixing cut-off transaction results

ericcorriel:
currently tx results are cut-off by the Send button. This is a quick fix to move the tx results on their own line. A better fix would be to change the html so the Send button is before the tx results in the DOM (which I tried (moving FunctionForm.jsx ln 235 to ln 178), but unfortunately just changing the order will make it so that the tx results do not show at all; more investigation is needed there).

Screen Shot 2022-10-13 at 05 16 02

Screen Shot 2022-10-13 at 04 49 39


codenamejason (changes requested):
@ericcorriel can we make it so the button is below the transaction result?

ericcorriel:
@codenamejason moving the button below the tx result could be done but I don't it would be desirable since it would leave a space between the input and the button by default.

Screenshot 2022-12-05 at 19 27 55

PR #914, opened 2022-10-14
better display of transaction results

ericcorriel:
Two things that I think would help people learn solidity better would be gaining an understanding of how gas works from the beginning. Scaffold.eth is a great iterative learning tool and this PR seeks to incorporate a more transparent view of how gas works into the framework.

There are two parts:
- New: display cost of transaction, in eth
- Improved: replace output of BigNumber hex values with values in wei/gwei/eth 

Screen Shot 2022-10-14 at 00 46 17
---
Screen Shot 2022-10-14 at 00 30 25

One note:
- I could not find where to set gas in the UI (hence I hardcoded it in this PR, which would obviously need to be replaced). Apologies if I'm totally missing it, but I think it would be helpful to be able to set gas (slider UI component) on the contracts page so as to be able to simulate transaction costs at different gas prices.


codenamejason (changes requested):
Looks good so far. Need to find a way to set the gas dynamically and not hardcoded before we merge.

PR #920, opened 2022-11-09
Added celo blockchain networks to scaffold eth

eliashezron:
I added celo blockchain networks to hardhat config file and react constants file to allow celo blockchain developers easily get started with scaffold-eth.


codenamejason (commented):
why all the changes to `YourContract.sol` this PR is more than just adding a network.

eliashezron:
It's because of prettier that just included double quotes to the contract.sol file when I saved it.

Check out the changes and you'll see that nothing was added to those files.

codenamejason:
@eliashezron `YourContract.sol` has a bunch of stuff added 🤔

eliashezron:
Let me do another PR

PR #926, opened 2022-11-29
Decouple ipfs storage and add support for Arwave via Akord

tmicltw:
Hi, and thanks for the great project!

I added support for Arweave to the NFT minting process so that users can choose whether they want to upload their files to IPFS or Arweave.
Although I added it to the `simple-nft-example` branch, I think that this could be more widely applicable to other parts of the projects as an alternative and more permanent storage solution to IPFS.
Uploading directly to Arweave requires AR tokens, which is not very user-friendly for newcomers, so I used Akord (which I am a member of) which provides 500MB of free Arweave storage.

Thanks a lot and let me know if you have any comments!

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,9 +36,9 @@ yarn deploy --network NETWORK_OF_CHOICE (localhost, kovan, rinkeby, mainnet)
 ![nft1](https://user-images.githubusercontent.com/526558/124386962-37e5dd00-dcb3-11eb-911e-0afce760d7ee.png)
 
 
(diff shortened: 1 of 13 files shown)


codenamejason:
@tmicltw can you add this to the `.example.env` 

```
REACT_APP_INFURA_PROJECT_ID=
REACT_APP_INFURA_PROJECT_SECRET=

AKORD_EMAIL=
AKORD_PASSWORD=
AKORD_VAULT_NAME=
```

codenamejason:
I created a new branch for this @tmicltw `simple-nft-example-arweave-ipfs`

PR #928, opened 2022-12-06
Use Docker compose

ferric-sol:
Updating scaffold-eth to use docker-compose

- Adds Dockerfile
- Adds docker-compose.yaml to start the chain and the frontend and expose the correct ports
- Point hardhat.config.js and constants.js to the container name (not sure if needed, perhaps better off in .env?)
- Update README


packages/react-app/src/constants.js
```diff
@@ -19,10 +19,16 @@ Note: this is not used when you are in the local hardhat chain.
...
+let localRpcUrl = process.env.REACT_APP_CODESPACES
   ? `https://${window.location.hostname.replace("3000", "8545")}`
   : "http://" + (global.window ? window.location.hostname : "localhost") + ":8545";
 
+// When running inside docker, we need to use the HARDHAT_HOST env variable
+// in order to access the container running hardhat
+if (process.env.HARDHAT_HOST) {
+  localRpcUrl = `http://${process.env.HARDHAT_HOST}:8545`;
+}
+
```
supriyaamisshra, line 31: This can be skipped I think because the container ports are exposed on the same network

supriyaamisshra (dismissed):
LGTM

packages/react-app/src/helpers/Web3ModalSetup.js
```diff
@@ -33,7 +33,7 @@ const web3ModalSetup = () =>
             10: "https://mainnet.optimism.io", // xDai
             100: "https://rpc.gnosischain.com", // xDai
             137: "https://polygon-rpc.com",
-            31337: "http://localhost:8545",
+            31337: `http://${process.env.HARDHAT_HOST || "localhost"}:8545`,
```
codenamejason, line 36: @ferric-sol I am going to add an export to `constants.js` with the name of `HARDHAT_HOST` so if you could update to use the new import like:

```javascript
import { ALCHEMY_KEY, INFURA_ID, HARDHAT_HOST } from "../constants";
```
on line 6 in `Web3ModalSetup.js`

and then update this line to reflect the new import
```javascript
31337: `http://${HARDHAT_HOST || "localhost"}:8545`,
```

🚨 see PR #955

ferric-sol: cool, done!

ferric-sol: guess i could remove || localhost as well, right? @codenamejason

README.md
```diff
@@ -41,6 +42,31 @@ cd scaffold-eth
...
+
+Now you have two containers, one running the frontend and one running hardhat
+
+Deploy your contract using:
+
+```bash
+cd scaffold-eth
+yarn deploy
+```
+
```
codenamejason, line 69: @ferric-sol should we still add the `yarn start` instructions here also? WDYT?

packages/react-app/src/constants.js
ferric-sol: wow, I think I may not need HARDHAT_HOST at all? I thought because app and chain were running in different containers I'd need to set the host for them to talk to each other. But if localhost works then HARDHAT_HOST isn't need at all, right?

Thanks for the review, btw!

PR #936, opened 2022-12-08
Enable CORS so safeApp manifest can be read by Gnosis App

santteegt:
The following files were added to enable CORS when using a SafeApp for development & production:

To enable CORS In development mode
* packages/react-app/src/setupProxy.js

To enable CORS when deploying to Surge:
* packages/react-app/public/CORS


codenamejason (approved):
This looks good!!
Screenshot 2022-12-27 at 10 20 55 PM
Screenshot 2022-12-27 at 10 21 53 PM
Screenshot 2022-12-27 at 10 22 06 PM

codenamejason:
Also see PR #889

damianmarti (changes requested):
Nice feature!!

Maybe we can add some comment at the README or somewhere about how to enable CORS.

cc/ @codenamejason where do you think is better to add this comment?

PR #945, opened 2022-12-15
line 4: add license field and lines: 91 & 94 fix openssl error

Bec0l:
Just suggesting some changes that fixed some errors for me


codenamejason:
@Bec0l thank you, we will address them in another PR #953

PR #952, opened 2022-12-28
Use local gas price hook using etherscan api

damianmarti:
https://docs.etherscan.io/api-endpoints/gas-tracker#get-gas-oracle


austintgriffith (approved):
🔥

PR #956, opened 2022-12-29
feat: upgradeable contracts starter kit, complete with upgrade-safe c…

xBalbinus:
Purpose:

This PR introduces a new starter kit for upgradeable contracts using the OpenZeppelin Upgrades API (since there is no starter kits right now for this yet, it seems).

Includes:
- Upgrade-safe contract (UUPS)
- Upgradeable contract deploy script
- Updated readme instructions to get contracts deployed, ready for frontend
- Script for upgrading function implementation

Testing:
- Contracts wise, trivial modifications were made to the original scaffold-eth contract
- Otherwise, all upgrade-related items (deployment, upgrade script) was tested by following the readme instructions from start to finish on localhost

Suggestion:
- To include this inside a new branch of this repo, or the scaffold eth examples repo.
- Feedback welcome, wondering if we should also do a Typescript version of this (wouldn't be too difficult)


codenamejason:
@xBalbinus I like this! 

Want to make sure it is introduced properly to the community and how we show the new folks how to use SE 🏗️

Let's coordinate on this and figure out the best way to introduce 🙌

damianmarti:
Great new feature!! 

But I think that is best to push it to a new branch. I think that is better to keep the main branch as simple as we can, and then you have the branches with a lot of other examples if you need something else.

codenamejason (approved):

PR #998, opened 2023-04-18
let's start using se2

austintgriffith:


DanieleSalatti (approved):
I would make it more prominent maybe?

Other than that, LGTM!

